PR #6, opened 2017-04-24
Add NSIS installer

MTRNord:
This PR should provide a setup.exe for Windows Users which is generated at appveyor. 

Questions to @mujx:
Nsis needs some fields which I am currently unclear what I need to fill in there. Here is a List of these:

- [x] Version
- [x] Company Name


win_build/nheko.nsi
```diff
@@ -0,0 +1,262 @@
+############################################################################################
```
mujx: Better put this file into `.ci/windows/nheko.nsi`

MTRNord:
I will change the things you pointed out later this day :)

PR #8, opened 2017-04-24
Change windows build to Release

walle303:
Visual Studio doesn't use `-DCMAKE_BUILD_TYPE` but uses `--config` instead


mujx:
Thanks a lot! Just squash the commits and I can merge it.

PR #10, opened 2017-04-30
fix various macOS specific problems

ara4n:
surely the maximum size for the text should be the width of the parent
(i.e. topLayout) minus the width of the avatar layout, which is max_height_?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ $ brew update
 $ brew install qt5
 ```
 
+N.B. you will need to pass `-DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.8.0_2`
+to cmake to point it at your qt5 install (tweaking the path as needed)
+
 ### Building
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -13,7 +13,8 @@ but you can of course receive and send messages in the rooms that you are a memb
 
 ### Requirements
 
-- Qt5 (5.8 is used for development).
+- Qt5 (5.7 or greater). Qt 5.7 adds support for color font rendering with
+  Freetype, which is essential to properly support emoji.
 - CMake 3.1 or greater.
 - A compiler that supports C++11.
     - Clang 3.3 (or greater).
@@ -22,16 +23,16 @@ but you can of course receive and send messages in the rooms that you are a memb
 ##### Arch Linux
 
 ```bash
-$ sudo pacman -S qt5-base cmake gcc
+$ sudo pacman -S qt5-base cmake gcc fontconfig
 ```
 
 ##### Ubuntu 14.04
 
 ```bash
-$ sudo add-apt-repository ppa:beineri/opt-qt562-trusty
+$ sudo add-apt-repository ppa:beineri/opt-qt58-trusty
 $ sudo add-apt-repository ppa:george-edison55/cmake-3.x
 $ sudo apt-get update
-$ sudo apt-get install qt56base cmake
+$ sudo apt-get install qt58base cmake
 ```
 
 ##### OSX (Xcode 7 or later)
--- a/src/ImageOverlayDialog.cc
+++ b/src/ImageOverlayDialog.cc
@@ -36,8 +36,6 @@ ImageOverlayDialog::ImageOverlayDialog(QPixmap image, QWidget *parent)
 
 	setWindowState(Qt::WindowFullScreen);
 
-	raise();
-
 	connect(this, SIGNAL(closing()), this, SLOT(closeDialog()));
 }
 
--- a/src/ImageOverlayDialog.cc
+++ b/src/ImageOverlayDialog.cc
@@ -42,16 +42,9 @@ ImageOverlayDialog::ImageOverlayDialog(QPixmap image, QWidget *parent)
 void ImageOverlayDialog::reject()
 {
 	// needed on macOS to recover the main menu after the dialog is closed(!)
+	// also affects KDE/Plasma.  XXX: There may be a better way of resetting the
+	// window state than this...
 	setWindowState(Qt::WindowNoState);
-
-	// It might be slightly cleaner to do:
-	//
-	// #ifdef Q_WS_MACX
-	// #include <Carbon/Carbon.h>
-	// ShowMenuBar();
-	// #endif
-	//
-	// ...but rather than being platform-specific, let's try to fix it at the Qt layer.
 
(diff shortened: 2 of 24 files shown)


src/RoomInfoListItem.cc
```diff
@@ -77,12 +77,12 @@ RoomInfoListItem::RoomInfoListItem(RoomInfo info, QWidget *parent)
 	avatarLayout_->addWidget(roomAvatar_);
 
 	roomName_ = new QLabel(info_.name(), textWidget_);
-	roomName_->setMaximumSize(parent->width() - 10, 20);
+	roomName_->setMaximumSize(parent->width() - max_height_, 20);
```
mujx: I don't think those are necessary, it just squashes the labels further inside. The 
```
setMaximumSize(parent->width(), max_height_);
``` 
line, was probably the bugfix. Could you confirm on this?

src/ImageOverlayDialog.cc
```diff
@@ -38,6 +38,23 @@ ImageOverlayDialog::ImageOverlayDialog(QPixmap image, QWidget *parent)
...
 
+void ImageOverlayDialog::reject()
+{
+	// needed on macOS to recover the main menu after the dialog is closed(!)
+	setWindowState(Qt::WindowNoState);
+
+	// It might be slightly cleaner to do:
+	//
+	// #ifdef Q_WS_MACX
+	// #include <Carbon/Carbon.h>
```
mujx: Will this work by default on OSX? If yes we can uncomment this section. Btw how other platforms would be affected by this?

ara4n: The problem on macOS here is that for whatever reason, when the ImageOverlayDialog is dismissed, the fullscreen state isn't fully reset for the app - meaning you lose the application menu bar for the rest of the app's lifetime.

I assume this is either a bug in Qt where it fails to call the MacOS API `ShowMenuBar()` to reset the UI when closing a QWindow with WindowState `QWindow::FullScreen`, or there's some missing housekeeping that lets Qt reset the app correctly.

I think the second option is more likely, so I'm faking the cleanup by forcing the QDialog to de-fullscreen itself before it gets destroyed, which correctly resets the app's menu bar.  This could also affect other platforms where Qt implements some application-wide behaviour for fullscreen (e.g. hiding the menu bar), so I think this is the better solution than doing the macOS-specific hack of breaking the Qt abstraction and directly calling the Carbon API needed to hack around the problem.

By the way, why does the constructor `raise()`?

src/RoomInfoListItem.cc
ara4n: I reverted the `roomName`, `roomTopic` and `setElidedText` sizing changes, which caused everything to break again:

screen shot 2017-04-30 at 12 42 46

I then re-added the setElidedText change, just to check if it was just the eliding operation which was mis-sized, which improved things a bit but still broke with rooms with a long room name (Megolm test, in this example):

screen shot 2017-04-30 at 12 44 37

...so I'm pretty sure that the patch here is correct.  My rationale is that `parent->width()` is returning the width of the whole `RoomInfoListItem`, so you should be sizing the width of the `roomName` and `roomTopic` boxes to that width, minus the width of the avatar (which is set to `max_height_`).

Then again, I know nothing about Qt :D

src/ImageOverlayDialog.cc
mujx: The solution is good enough for me. Since I'm pretty new to Qt myself, I might have missing something or the full screen widget isn't implemented properly. So maybe it's not a Qt bug but we have missed the proper API calls.

You can remove `raise()`, it's probably a leftover from previous iterations.

src/RoomInfoListItem.cc
mujx: Yes you're right. But on Linux the situation is a little different. The room topic is cut off a lot sooner.

room-list


The problem with the room name is that the font size isn't taken into consideration. Same goes for the room topic. We should use `QFontMetrics` to cut off the text. Maybe the screen resolution has an impact on this and this was the problem all along.

test

mujx:
I'm a little confused with the last three commits. Probably something went wrong with the rebase because `raise()` is not removed.

The `parent->width() - max_height_` is correct. I just added a `-5` on my branch so the text doesn't touch the right border.  The only thing missing from the PR is to replace the stylesheet font size as shown in `fix/roomlist-text`. You can also revert the `setMaximumSize` as it keeps the widget smaller (there are examples of this on the previous comments).

jansol:
> I just added a -5 on my branch so the text doesn't touch the right border.

Please make that a const variable (`PADDING_RIGHT` or so)

ara4n:
I added raise() back in when merging because you hadn't removed it when adding the close hook, and because i realised i had assumed raise() was a python-style "raise a Qt exception" rather than the "raise this window to the top"... which seems like a reasonable thing to do.  Happy to kill it again though if that's the right thing to do :)

I reverted to allow for a clean merge with your branch.

i'll test to see if reverting the setMaximumSize is ok on OSX - but be aware that your fix-room-list branch does have nasty 4px of horizontal overscroll when i tested it earlier.

-- 
Matthew Hodgson
Matrix.org

mujx:
Thanks for clearing that up. I thought you were going to integrate the font part in your branch since most of the work has already been done there. Could you post a screenshot with the 4px problem?

When you are ready just squash the commits.

ara4n:
(out-of-band it seems the 4px problem is indeed the need for setMaximumSize on the RoomInfoListItem)

Slightly dubious about squashing the commits (why not just do a squash merge?), but I guess i can do so on the assumption that nobody else is playing with this branch...

ara4n:
okay, i've tried to rebase and squash the null commits, although it looks to have made a bit of a mess :( hopefully it will merge simply though.

mujx:
Could you squash the commits into one? Btw the issue with the overlay image appearing tiny also happens on KDE/Plasma.

README.md
```diff
@@ -42,6 +42,9 @@ $ brew update
 $ brew install qt5
 ```
 
+N.B. you will need to pass `-DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.8.0_2`
```
mujx: Did you try the path on travis.yml? `CMAKE_PREFIX_PATH=/usr/local/opt/qt5`

ara4n:
ok, this is a complete mess now - opened #15 to replace it.

PR #35, opened 2017-05-26
Improve login flow

jansol:
* Validate both inferred and explicitly entered server addresses by attempting to call the /versions endpoint
* If the domain from the mxid fails validation, try prefixing it with 'matrix'
* Only show server address field if address validation ultimately fails
* Show shiny status indicators next to the mxid or server field


jansol:
Travis failures look unrelated.

src/LoginPage.cc
```diff
@@ -53,13 +47,6 @@ LoginPage::LoginPage(QSharedPointer<MatrixClient> client, QWidget *parent)
 	QIcon advanced_settings_icon;
 	advanced_settings_icon.addFile(":/icons/icons/cog.png", QSize(), QIcon::Normal, QIcon::Off);
 
-	advanced_settings_button_->setIcon(advanced_settings_icon);
-	advanced_settings_button_->setIconSize(QSize(24, 24));
-
-	top_bar_layout_->addWidget(back_button_, 0, Qt::AlignLeft | Qt::AlignVCenter);
-	top_bar_layout_->addStretch(1);
```
jansol: Which two of these four?

src/LoginPage.cc
```diff
@@ -142,63 +177,111 @@ void LoginPage::loginError(QString error)
 	error_label_->setText(error);
 }
 
-void LoginPage::onLoginButtonClicked()
+void LoginPage::onMatrixIdEntered()
 {
+	guessing_server = true;
```
jansol: Can we at least agree on a naming convention for variables first?

src/LoginPage.cc
```diff
@@ -142,63 +177,111 @@ void LoginPage::loginError(QString error)
...
-		QString password = password_input_->text();
+		matrixid_success_->hide();
+		matrixid_spinner_->show();
+	}
 
-		QString home_server = custom_domain_.isEmpty()
-					      ? matrixid_input_->text().split(":").at(1)
-					      : custom_domain_;
+	server_layout_->invalidate();
+	matrixid_layout_->invalidate();
```
jansol: The layouts wouldn't realign themselves when hiding things otherwise. That might be different if the hidden elements are actually removed from the layout.

src/LoginPage.cc
```diff
@@ -142,63 +177,111 @@ void LoginPage::loginError(QString error)
...
 	password_input_->clear();
+	server_input_->clear();
 
-	if (settings_modal_ != nullptr) {
-		settings_modal_->deleteLater();
-		settings_modal_ = nullptr;
-	}
+	server_success_->hide();
+	server_error_->hide();
+	server_spinner_->hide();
```
jansol: Having UI elements disappear is terribly confusing, especially when combined with not showing the 'ok' icon.

src/LoginPage.cc
mujx: The `top_bar_layout_` lines.

src/LoginPage.cc
mujx: `reset()` is called when we're done with the login page and there is no reason keeping optional elements visible. For example if the user calls logout and tries to login again he will see the server input.

src/LoginPage.cc
mujx: You're right, I started with snake case for member variables but camel case seems to be a 
 better option for longer names.

mujx:
Thanks a lot for your work!

PR #43, opened 2017-07-04
Fix emoji alignment issue

george99g:
Without the flag, the emoji were being drawn off-center, leading to them getting slightly clipped in some places or going out of bounds in others. It isn't too obvious right now, but it becomes obvious if you add a background on hover.
Here are some screenshots to show the difference:
image
image


mujx:
Good catch! Thanks.

george99g:
You're welcome!

PR #67, opened 2017-09-23
allow handling of incomplete /sync responses

krombel:
matrix-org/synapse#2358 proposes a change to only contain objects
in the sync stream where values get updated.
The current behavior leads nheko to fail when that is used.

This PR makes nheko compatible to that.


mujx:
Thanks for being proactive on this one! Looks good to me. 

Could you add a few test cases for the sync parsing to show that it doesn't fail?

PR #69, opened 2017-09-23
Ctrl-Q from the main window quits the application

### LINKED ISSUE #61: Allow Ctrl+Q to quit
A convenience shortcut which every program ever seems to support. Use Ctrl+Q to quit the program

vurpo:
Fixes #61.

I've only tested this on my laptop running Fedora, but it seems simple enough.
Does macOS and its Command key need to be separately handled?


Ralith:
Is there some reason not to use the more portable and configurable `QKeySequence::Quit` instead?

vurpo:
That's a good question.

vurpo:
Now it does (tested and still works the same)

mujx:
Thanks! Btw cmd+Q already worked on macOS.

PR #76, opened 2017-09-28
Remove replaceEmoji

vberger:
This remove the `replaceEmoji` function that didn't properly handle UTF-16 contents of QString.

I think this new version should still allow the proper recognition of emojis, at least it works on my computer. It'd be nice if some other people could test it as well.

### DIFF
--- a/include/TimelineItem.h
+++ b/include/TimelineItem.h
@@ -80,8 +80,6 @@ private:
         void setupAvatarLayout(const QString &userName);
         void setupSimpleLayout();
 
-        QString replaceEmoji(const QString &body);
-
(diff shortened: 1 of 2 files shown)


mujx:
Although it should, this doesn't work on my system. That was my implementation before switching to that `replaceEmoji` function. 

I made it work without `replaceEmoji`, if you set the QLabel's font with `body_->setFont(QFont("Open Sans, Emoji One")`. But it still has some issues with the font size.

vberger:
Ok. I have absolutely no idea why this does not work on your computer, or why the current implementation crashes on my computer.

But if this only works for me, then the issue is more likely with my system.

PR #82, opened 2017-10-01
Fixes #50.

### LINKED ISSUE #50: Missing tray icon in KDE

rokups:
On KDE desktop icon failed to appear because TrayIcon requested a zero-size rect. Implementing MsgCountComposedIcon::availableSizes() method fixes that.

After icon became visible it was not transparent, and places that should have been transparent contained artifacts likely due to uninitialized memory. Implementing MsgCountComposedIcon::pixmap() which returns a pixmap with alpha channel fixes that.


mujx:
That works perfectly. Thanks!

PR #91, opened 2017-10-09
Fixed Version in desktop file.

xvitaly:
**Version** must contains version of desktop file schema, not version of application.


xvitaly:
More information about desktop files available [here](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s05.html).

PR #96, opened 2017-10-20
Fix sidebar resize

### LINKED ISSUE #95: Weird sidebar movement in some rooms
Sidebar appears to have multiple issues with handling movement.

* Some rooms work perfectly fine.
* Other rooms appear to work ok, but sidebar can not move as much to the right as in other rooms.
* Some rooms have sidebar "fixed", it can not be moved freely to the left or right. Splitter being locked into position causes sidebar width changes when switching between rooms where sidebar works and where it is locked.
* In rooms where sidebar is locked collapsing it to the left causes incorrect display of entire application window. Releasing splitter when sidebar is collapsed in this state leaves application in broken state and restart is required.

See video below:
[](https://www.youtube.com/watch?v=VlLRQBAFM4Y)

rokups:
Fixes #95 

Your previous commit did the trick. I made label to take (almost) entire width and expand/contract on left click. Resizing of sidebar now works equally good with short and long room topics.

Also fixed html formatting of elided text - html regex has to be applied after elision, otherwise label would attempt to display invalid html and spam console with errors. There is one problem remaining - if url is partially visible then clicking it would open url consisting of just visible part. It will do for now i guess.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,7 @@ set(SRC_FILES
     src/ui/Badge.cc
     src/ui/LoadingIndicator.cc
     src/ui/FlatButton.cc
+    src/ui/Label.cc
     src/ui/OverlayModal.cc
     src/ui/ScrollBar.cc
     src/ui/SnackBar.cc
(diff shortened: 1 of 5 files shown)


mujx:
Nice work on the label click! I'm ok with the partial link problem you mentioned, we can iterate later.

Could you run `make lint` or just clang-format (version 5) on the files you changed to fix the build?

rokups:
Done. Will keep it in mind for future contributions ;)

PR #104, opened 2017-10-30
Use system color scheme (using a Qt stylesheet)

vurpo:
With this PR, the color scheme of nheko obeys the default color theme of Qt (i.e. the system theme). It uses a Qt stylesheet to accomplish this, which means replacing the color theme with a custom theme would only be a matter of writing a new style sheet and loading it into the app.

### DIFF
--- a/include/RoomInfoListItem.h
+++ b/include/RoomInfoListItem.h
@@ -47,8 +47,10 @@ class RoomInfoListItem : public QWidget
         Q_PROPERTY(QColor titleColor READ titleColor WRITE setTitleColor)
         Q_PROPERTY(QColor subtitleColor READ subtitleColor WRITE setSubtitleColor)
 
-        Q_PROPERTY(QColor highlightedTitleColor READ highlightedTitleColor WRITE setHighlightedTitleColor)
-        Q_PROPERTY(QColor highlightedSubtitleColor READ highlightedSubtitleColor WRITE setHighlightedSubtitleColor)
+        Q_PROPERTY(
+          QColor highlightedTitleColor READ highlightedTitleColor WRITE setHighlightedTitleColor)
+        Q_PROPERTY(QColor highlightedSubtitleColor READ highlightedSubtitleColor WRITE
+                     setHighlightedSubtitleColor)
 
 public:
         RoomInfoListItem(QSharedPointer<RoomSettings> settings,
@@ -87,7 +89,10 @@ public:
         inline void setBackgroundColor(QColor &color) { backgroundColor_ = color; }
 
         inline void setHighlightedTitleColor(QColor &color) { highlightedTitleColor_ = color; }
-        inline void setHighlightedSubtitleColor(QColor &color) { highlightedSubtitleColor_ = color; }
+        inline void setHighlightedSubtitleColor(QColor &color)
+        {
+                highlightedSubtitleColor_ = color;
+        }
 
         inline void setTitleColor(QColor &color) { titleColor_ = color; }
         inline void setSubtitleColor(QColor &color) { subtitleColor_ = color; }
--- a/include/RoomInfoListItem.h
+++ b/include/RoomInfoListItem.h
@@ -38,6 +38,15 @@ struct DescInfo
 class RoomInfoListItem : public QWidget
 {
         Q_OBJECT
+        Q_PROPERTY(QColor highlightedBackgroundColor
+                   READ   highlightedBackgroundColor
+                   WRITE  setHighlightedBackgroundColor)
+        Q_PROPERTY(QColor hoverBackgroundColor
+                   READ   hoverBackgroundColor
+                   WRITE  setHoverBackgroundColor)
+        Q_PROPERTY(QColor backgroundColor
+                   READ   backgroundColor
+                   WRITE  setBackgroundColor)
 
 public:
         RoomInfoListItem(QSharedPointer<RoomSettings> settings,
@@ -51,12 +60,20 @@ public:
         void clearUnreadMessageCount();
         void setState(const RoomState &state);
 
-        bool isPressed() const { return isPressed_; };
-        RoomState state() const { return state_; };
-        int unreadMessageCount() const { return unreadMsgCount_; };
+        bool isPressed() const { return isPressed_; }
+        RoomState state() const { return state_; }
+        int unreadMessageCount() const { return unreadMsgCount_; }
 
         void setAvatar(const QImage &avatar_image);
         void setDescriptionMessage(const DescInfo &info);
+
+        inline QColor highlightedBackgroundColor() const { return highlightedBackgroundColor_; }
+        inline QColor hoverBackgroundColor() const { return hoverBackgroundColor_; }
+        inline QColor backgroundColor() const { return backgroundColor_; }
+
+        inline void setHighlightedBackgroundColor(QColor &color) { highlightedBackgroundColor_ = color; }
+        inline void setHoverBackgroundColor(QColor &color) { hoverBackgroundColor_ = color; }
+        inline void setBackgroundColor(QColor &color) { backgroundColor_ = color; }
 
 signals:
         void clicked(const QString &room_id);
@@ -98,4 +115,8 @@ private:
 
         int maxHeight_;
         int unreadMsgCount_ = 0;
+
+        QColor highlightedBackgroundColor_;
+        QColor hoverBackgroundColor_;
+        QColor backgroundColor_;
 };
--- a/include/RoomInfoListItem.h
+++ b/include/RoomInfoListItem.h
@@ -44,6 +44,12 @@ class RoomInfoListItem : public QWidget
           QColor hoverBackgroundColor READ hoverBackgroundColor WRITE setHoverBackgroundColor)
         Q_PROPERTY(QColor backgroundColor READ backgroundColor WRITE setBackgroundColor)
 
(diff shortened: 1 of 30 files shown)


mujx:
Can we set up the stylesheet to have the current theme, cause on mac for example the colors are broken (haven't checked on windows yet). 

So the default theme will remain the same and we will have an option (will be added later) to load the desired theme.

vurpo:
Can you send a screenshot from macOS? I don't have a Mac to test on.

vurpo:
I can test on Windows.

src/EmojiPanel.cc
```diff
@@ -55,7 +53,7 @@ EmojiPanel::EmojiPanel(QWidget *parent)
         contentLayout->setMargin(0);
 
         auto emojiCategories = new QFrame(mainWidget);
-        emojiCategories->setStyleSheet("background-color: #f2f2f2");
+        // emojiCategories->setStyleSheet("background-color: #f2f2f2");
```
mujx: The emoji categories would need a different background color from the rest of the panel.

src/RoomInfoListItem.cc
```diff
@@ -96,29 +96,24 @@ RoomInfoListItem::paintEvent(QPaintEvent *event)
...
-        p.setPen(QColor("#333"));
+        // p.setPen(QColor("#333"));
 
         QRect avatarRegion(Padding, Padding, IconSize, IconSize);
 
         // Description line with the default font.
         int bottom_y = maxHeight_ - Padding - Padding / 3 - metrics.ascent() / 2;
 
         if (width() > ui::sidebar::SmallSize) {
-                if (isPressed_) {
```
mujx: We would need to keep those lines using a non hardcoded color.

src/RoomInfoListItem.cc
```diff
@@ -130,11 +125,6 @@ RoomInfoListItem::paintEvent(QPaintEvent *event)
                   state_.getName(), Qt::ElideRight, (width() - IconSize - 2 * Padding) * 0.8);
                 p.drawText(QPoint(2 * Padding + IconSize, top_y), name);
 
-                if (!isPressed_) {
```
mujx: We would also need to keep those.

src/TextInputWidget.cc
```diff
@@ -86,6 +85,7 @@ TextInputWidget::TextInputWidget(QWidget *parent)
         spinner_ = new LoadingIndicator(this);
         spinner_->setFixedHeight(32);
         spinner_->setFixedWidth(32);
+        spinner_->setColor(palette().color(QPalette::Text));
```
mujx: Spinner should have its own color instead of using `QPalette::Text`.

src/TextInputWidget.cc
```diff
@@ -96,7 +96,7 @@ TextInputWidget::TextInputWidget(QWidget *parent)
         input_->setFont(font);
         input_->setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
         input_->setPlaceholderText(tr("Write a message..."));
-        input_->setStyleSheet("color: #333333; border: none; padding-top: 5px; margin: 0 5px");
+        input_->setStyleSheet("border: none; padding-top: 5px; background-color: rgba(0,0,0,0);");
```
mujx: This makes the context menu of the input transparent.

src/TimelineItem.cc
```diff
@@ -305,7 +305,7 @@ TimelineItem::TimelineItem(const events::MessageEvent<msgs::Text> &event,
 void
 TimelineItem::generateBody(const QString &body)
 {
-        QString content("<span style=\"color: black;\"> %1 </span>");
+        QString content("<span> %1 </span>");
```
mujx: I think you can remove those `span` tags if there is no color defined.

src/LoginPage.cc
```diff
@@ -136,7 +136,7 @@ LoginPage::LoginPage(QSharedPointer<MatrixClient> client, QWidget *parent)
 
         error_label_ = new QLabel(this);
         error_label_->setFont(font);
-        error_label_->setStyleSheet("color: #E22826");
+        //        error_label_->setStyleSheet("color: #E22826");
```
mujx: We would need a red-ish color for the error label.

src/RegisterPage.cc
```diff
@@ -112,11 +112,11 @@ RegisterPage::RegisterPage(QSharedPointer<MatrixClient> client, QWidget *parent)
 
         error_label_ = new QLabel(this);
         error_label_->setFont(font);
-        error_label_->setStyleSheet("color: #E22826");
+        //        error_label_->setStyleSheet("color: #E22826");
```
mujx: Same as the login page.

src/TimelineItem.cc
```diff
@@ -30,7 +30,7 @@
 #include "TimelineViewManager.h"
 
 static const QRegExp URL_REGEX("((?:https?|ftp)://\\S+)");
-static const QString URL_HTML = "<a href=\"\\1\" style=\"color: #333333\">\\1</a>";
+static const QString URL_HTML = "<a href=\"\\1\">\\1</a>";
```
mujx: Probably we need to specify a color for the links also. The default blue might be too hard for some themes.

src/TimelineItem.cc
```diff
@@ -396,8 +396,8 @@ TimelineItem::setupAvatarLayout(const QString &userName)
 
         userAvatar_ = new Avatar(this);
         userAvatar_->setLetter(QChar(userName[0]).toUpper());
-        userAvatar_->setBackgroundColor(QColor("#eee"));
-        userAvatar_->setTextColor(QColor("black"));
+        //        userAvatar_->setBackgroundColor(QColor("#eee"));
```
mujx: We also need a color for the background, otherwise it's the same as the timeline and only a letter is visible.

resources/styles/nheko.qss
```diff
@@ -0,0 +1,40 @@
+* {
+        background-color: white;
+        color: black;
+}
+
+QLabel {
+        color: black;
```
mujx: `color: #333`

resources/styles/nheko.qss
```diff
@@ -0,0 +1,40 @@
+* {
+        background-color: white;
+        color: black;
```
mujx: `color: #333`

resources/styles/nheko.qss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+RoomInfoListItem {
+        qproperty-highlightedBackgroundColor: #38A3D8;
+        qproperty-hoverBackgroundColor: rgba(200,200,200,128);
+        qproperty-backgroundColor: white;
+}
+
+LoadingIndicator {
+        /* qproperty-color: ; */
```
mujx: `qproperty-color: #acc7dc`

vurpo: This one needs to get two cases (I'll add an object name to one of them and deal with it that way), because the same spinner widget is also used on the file upload button

mujx: Couldn't be done within the `LoadingIndicator` class directly?

vurpo: No, because there are two instances of `LoadingIndicator` within the app, that have to get different colors. Anyway, the solution is simple: I will `spinner_->setObjectName("ChatPageLoadSpinner");` for the chat page loading spinner, and then I can just put `#ChatPageLoadSpinner { qproperty-color: #acc7dc; }` in the stylesheet (and the same for the other spinner, with a different name). This way, different styles can be applied to different instances of the same widget.

src/TimelineItem.cc
vurpo: I found that the link color becomes a reasonable one chosen by the theme if this is not set. Additionally, it was a pain trying to get links inside a `TimelineItem` to get a color set by the stylesheet; there is no way to set the color of links inside an existing `QLabel`, so you would have to generate the content of all the `QLabel`s again with the new color set for every link when the link color changes. And properties defined in the stylesheet are set only *after* the object's initializer is called, which means the `TimelineItem` can not yet know its preferred link color at the time that all the `QLabel`s are being created.

include/RoomInfoListItem.h
```diff
@@ -51,13 +62,36 @@ class RoomInfoListItem : public QWidget
...
-        RoomState state() const { return state_; };
-        int unreadMessageCount() const { return unreadMsgCount_; };
+        bool isPressed() const { return isPressed_; }
+        RoomState state() const { return state_; }
+        int unreadMessageCount() const { return unreadMsgCount_; }
 
         void setAvatar(const QImage &avatar_image);
         void setDescriptionMessage(const DescInfo &info);
 
+        inline QColor highlightedBackgroundColor() const { return highlightedBackgroundColor_; }
```
mujx: You can remove the `inline` keyword because it's implied.

resources/styles/nheko.qss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+QMenu, QMenu > * {
+        background-color: white;
+}
+
+FlatButton {
+        qproperty-foregroundColor: #333;
+}
+
```
mujx: We also need 
```
RaiseButton {
    qproperty-foregroundColor: white
}
```
which is the text for the buttons in the login page etc.

mujx:
Merged in https://github.com/mujx/nheko/commit/929b2df6fb28a1e431027afb032f5eeefd0106b7

PR #105, opened 2017-10-31
implemented sending of typing notifications

### LINKED ISSUE #88: Send and receive typing notifications
- [ ] **Send** 
- [x] **Receive** -  A good place to show them would be under the main text input field. The text could have the form of "*bob* and *alice* are typing ... " etc.

karroffel:
fixes #88

### DIFF
--- a/src/ChatPage.cc
+++ b/src/ChatPage.cc
@@ -122,6 +122,9 @@ ChatPage::ChatPage(QSharedPointer<MatrixClient> client, QWidget *parent)
         contentLayout_->addWidget(typingDisplay_);
         contentLayout_->addWidget(text_input_);
 
+        typingRefresher_ = new QTimer(this);
+        typingRefresher_->setInterval(TYPING_REFRESH_TIMEOUT);
+
         user_info_widget_ = new UserInfoWidget(sideBarTopWidget_);
         sideBarTopWidgetLayout_->addWidget(user_info_widget_);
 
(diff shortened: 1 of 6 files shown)


src/ChatPage.cc
```diff
@@ -139,6 +142,10 @@ ChatPage::ChatPage(QSharedPointer<MatrixClient> client, QWidget *parent)
 
                 typingDisplay_->setUsers(users);
         });
+        connect(room_list_, &RoomList::roomChanged, this, [=]() {
+                typingRefresher_->stop();
+                text_input_->stopTyping();
```
mujx: If I'm not mistaken `stopTyping` will emit the `stoppedTyping` event which is also handled here. Maybe we can just call `MatrixClient::sendTypingStoppedNotifications` here directly like the `stoppedTyping` handler does. The timer on the text input will expire anyway.

src/ChatPage.cc
```diff
@@ -159,6 +166,20 @@ ChatPage::ChatPage(QSharedPointer<MatrixClient> client, QWidget *parent)
                                 room_list_->updateUnreadMessageCount(roomid, count);
                 });
 
+        connect(text_input_, &TextInputWidget::startedTyping, this, [=]() {
+                typingRefresher_->start();
+                client->sendTypingNotification(current_room_);
```
mujx: `client_` is the class member. `client` will only work on the constructor.

PR #106, opened 2017-10-31
Communities

vurpo:
Adds a sidebar with each community that the user is a member of, Discord-style.

Some open questions still: should the looks of the sidebar be changed in any way still? Should the layout be changed in any way? Should the whole feature be always on or behind a configuration flag?


vurpo:
Also there's no caching of anything yet.

src/MatrixClient.cc
```diff
@@ -464,6 +464,45 @@ MatrixClient::getOwnProfile() noexcept
...
+                }
+
+                auto data = reply->readAll();
+                auto json = QJsonDocument::fromJson(data).object();
+                try {
+                        QList<QString> response;
+                        for (auto it = json["groups"].toArray().constBegin();
+                             it != json["groups"].toArray().constEnd();
+                             it++) {
+                                response.append(it->toString());
```
karroffel: I get a SEGV here when accessing `it.operator->()`. While debugging I found that the `...toArray().constEnd()` invalidates the iterator. When using a single reference to the array then it works fine.

PureTryOut:
Maybe it's a good idea to also updated the screenshot in the README?

mujx:
Is this ready for review?

vurpo:
No, partially because I haven't brought it up to date with the new changes yet, but also because I don't know what the Matrix team's intention is currently with tags and communities (@ara4n talked about this in the nheko room?)

Of course if you think the functionality is acceptable as it is then I think it can be updated and then merged, but I think it's a good idea to follow what's happening in Riot and Matrix overall.

ara4n:
honestly we’re experimenting on riot-web too. current intended behaviour there is i think very similar to this PR, although also adding the idea of “tagging a room to a group” - ie shoving “+foo:bar.com” as a tag name rather than “m.favourite” or whatever. This lets users adhoc associate random rooms with their view of a group they so desire: eg if they have a private group chat with folks in a given community and want to pin it to that view of a community despite the room otherwise not being technically related to that community. Alternatively we may ditch this and just use heuristics to show rooms in a given community based on who’s in them (if they are not explicitly members of a given group).

ara4n:
ie the more experiments and suggestions the better so we can try to find a generic UX that doesn’t suck :) https://docs.google.com/document/d/1eSKqH9NuYC2aBt8fOWQuICS84KfL2eNlIWRoV5DinoI are our notes for the riot design; currently on “New Strawman”...

vurpo:
Sorry about the confusion, I committed on a different branch now, and now my latest changes are on the `communities` branch instead of `communities-merged`.

PR #108, opened 2017-11-01
Implement a setting for the tray icon

JaniM:
Adds an option `user/window/tray`. If that option is set to false, nheko exits when the window is closed. Defaults to true.

Signed-off-by: Jani Mustonen <janijohannes@kapsi.fi>

### DIFF
--- a/include/UserSettingsPage.h
+++ b/include/UserSettingsPage.h
@@ -66,6 +66,7 @@ protected:
 
 signals:
         void moveBack();
+        void trayOptionChanged(const bool value);
 
 private:
         // Layouts
--- a/include/UserSettingsPage.h
+++ b/include/UserSettingsPage.h
@@ -66,7 +66,7 @@ protected:
 
 signals:
         void moveBack();
-        void trayOptionChanged(const bool value);
+        void trayOptionChanged(bool value);
 
 private:
         // Layouts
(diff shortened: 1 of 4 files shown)


include/UserSettingsPage.h
```diff
@@ -66,6 +66,7 @@ class UserSettingsPage : public QWidget
 
 signals:
         void moveBack();
+        void trayOptionChanged(const bool value);
```
mujx: No need for `const`.

src/MainWindow.cc
```diff
@@ -94,6 +94,10 @@ MainWindow::MainWindow(QWidget *parent)
                 pageStack_->setCurrentWidget(chat_page_);
         });
 
+        connect(userSettingsPage_, &UserSettingsPage::trayOptionChanged, this, [=](const bool value) {
```
mujx: You can connect it to `TrayIcon::setVisible` directly without the lambda.

src/MainWindow.cc
```diff
@@ -113,9 +117,7 @@ MainWindow::MainWindow(QWidget *parent)
 
         QSettings settings;
 
-        if (!settings.value("user/window/tray", true).toBool()) {
-                trayIcon_->hide();
-        }
+        trayIcon_->setVisible(settings.value("user/window/tray", true).toBool());
```
mujx: We should use the `userSettings_` object to retrieve user configuration so we don't have to define the settings path every time.

mujx:
Thanks!

PR #109, opened 2017-11-01
Improvements to the quick switcher

JaniM:
Added a simple disambiguation to the room switcher.
Fixed ctrl+k (it got grabbed by the text input).
Fixed keyboard focus after closing the quick switcher (now automatically focuses the text input).

### DIFF
--- a/src/ChatPage.cc
+++ b/src/ChatPage.cc
@@ -562,6 +562,7 @@ ChatPage::showQuickSwitcher()
                 connect(quickSwitcher_.data(), &QuickSwitcher::closing, this, [=]() {
                         if (!this->quickSwitcherModal_.isNull())
                                 this->quickSwitcherModal_->fadeOut();
+                        this->text_input_->setFocus(Qt::FocusReason::PopupFocusReason);
                 });
         }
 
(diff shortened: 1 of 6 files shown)


src/MainWindow.cc
```diff
@@ -125,10 +130,7 @@ MainWindow::MainWindow(QWidget *parent)
 void
 MainWindow::keyPressEvent(QKeyEvent *e)
```
mujx: You can remove this function.

src/ChatPage.cc
```diff
@@ -575,8 +576,12 @@ ChatPage::showQuickSwitcher()
 
         QMap<QString, QString> rooms;
 
-        for (auto it = state_manager_.constBegin(); it != state_manager_.constEnd(); ++it)
-                rooms.insert(it.value().getName(), it.key());
+        for (auto it = state_manager_.constBegin(); it != state_manager_.constEnd(); ++it) {
+                QString deambiguator = it.value().canonical_alias.content().alias();
```
mujx: That's ok for now but eventually we will have to show the room list in a more custom way.

JaniM: Yes of course. I just wanted to ensure that all rooms get listed as is.

PR #119, opened 2017-11-04
Add input history, enable multi-line input, refactor commands

Ralith:
This ports over the interesting bits of nachat's input field, minus completions, which are a larger project.

Fixes the transmission of mis-typed commands as messages, fixes inability to send messages that start with a command, and does some initial work towards automatically resizing the input field to fit
the input message.

Future work should involve making input widget state per-room instead of global and working out how to make the chat page layout actually respect the input widget's desire to dynamically resize itself.

### DIFF
--- a/include/TextInputWidget.h
+++ b/include/TextInputWidget.h
@@ -17,6 +17,8 @@
 
 #pragma once
 
+#include <deque>
+
 #include <QHBoxLayout>
 #include <QPaintEvent>
 #include <QTextEdit>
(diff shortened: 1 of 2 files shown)


include/TextInputWidget.h
```diff
@@ -29,26 +31,36 @@
...
         void startedTyping();
         void stoppedTyping();
+        void message(QString);
+        void command(QString name, QString args);
+
+protected:
+        void keyPressEvent(QKeyEvent *event) override;
+
+private:
+        std::deque<QString> true_history_, working_history_;
```
karroffel: Why not use a `QQueue`? It doesn't really matter, but as far as I can see, the rest of the codebase uses mostly Qt containers, not STL ones.

Ralith: Why use a nonstandard container when a standard one will do just fine?

karroffel: *C O N S I S T E N C Y* 😎 👉 👉

Ralith: Consistency with the rest of the world is among the advantages of standardization, yes :smile:

karroffel: I mean in the nheko codebase. If you search for `std::` in the whole source then you'll see that not a single container is used, mostly things like `std::max` or functions for randomization.

Using `std::queue` instead of `QQueue` is just a bit alien. As I said, it doesn't really matter, it's just different than everything else.

src/TextInputWidget.cc
```diff
@@ -65,6 +108,65 @@ FilteredTextEdit::stopTyping()
...
+        return size;
+}
+
+void
+FilteredTextEdit::submit()
+{
+        if (true_history_.size() == INPUT_HISTORY_SIZE)
+                true_history_.pop_back();
+        true_history_.push_front(toPlainText());
+        working_history_ = true_history_;
```
mujx: This causes the working history to falsely have 127 (mostly empty) entries. If a message is sent you can keep going up to the working history. 

It can be resolved by initializing `true_history_` to 0.

Ralith: Good catch, thanks! Not sure what I was thinking there.

Issue #132, opened 2017-11-15
[feature request] Support pasting images into room


aditya3098:
I would really like the ability to control-v images into chat, I am putting this out there in case its not too time consuming to implment, while I will also try to implement this myself

mujx:
Drag and drop files in general could also work.

Ralith:
Drag and drop is probably a simpler problem to solve, in fact.

PR #138, opened 2017-11-25
Implement #132: Support pasting images into room

christarazi:
Implements #132 to allow pasting of images using ctrl-v.

### DIFF
--- a/src/TextInputWidget.cc
+++ b/src/TextInputWidget.cc
@@ -66,11 +66,6 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
                 typingTimer_->start();
         }
 
-        if (event->matches(QKeySequence::Copy)) {
-                insertFromMimeData(QApplication::clipboard()->mimeData());
-                return;
-        }
-
         switch (event->key()) {
         case Qt::Key_Return:
         case Qt::Key_Enter:
--- a/src/TextInputWidget.cc
+++ b/src/TextInputWidget.cc
@@ -185,6 +185,14 @@ FilteredTextEdit::submit()
         history_index_ = 0;
 
         QString text = toPlainText();
+
+        // Remove image from plaintext so it doesn't show in message.
+        if (isImage_) {
+                text.remove(QChar{0xfffc}); // Unicode for OBJECT REPLACEMENT CHARACTER
+                emit image(pastedImagePath_);
+                isImage_ = false;
+        }
+
         if (text.startsWith('/')) {
                 int command_end = text.indexOf(' ');
                 if (command_end == -1)
@@ -196,9 +204,6 @@ FilteredTextEdit::submit()
                 } else {
                         command(name, args);
                 }
-        } else if (isImage_) {
-                emit image(pastedImagePath_);
-                isImage_ = false;
         } else {
                 message(std::move(text));
         }
--- a/src/TextInputWidget.cc
+++ b/src/TextInputWidget.cc
@@ -66,6 +66,11 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
                 typingTimer_->start();
         }
 
+        if (event->matches(QKeySequence::Copy)) {
+                insertFromMimeData(QApplication::clipboard()->mimeData());
+                return;
+        }
+
         switch (event->key()) {
         case Qt::Key_Return:
         case Qt::Key_Enter:
@@ -94,12 +99,6 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
                         --history_index_;
                         setPlainText(working_history_[history_index_]);
                         moveCursor(QTextCursor::End);
-                }
-                break;
-        }
-        case Qt::Key_V: {
-                if (event->modifiers() == Qt::ControlModifier) {
-                        insertFromMimeData(QApplication::clipboard()->mimeData());
                 }
                 break;
         }
--- a/src/TextInputWidget.cc
+++ b/src/TextInputWidget.cc
@@ -16,10 +16,13 @@
  */
 
 #include <QAbstractTextDocumentLayout>
+#include <QApplication>
+#include <QClipboard>
 #include <QDebug>
 #include <QFile>
 #include <QFileDialog>
 #include <QImageReader>
+#include <QMimeData>
 #include <QPainter>
 #include <QStyleOption>
 
@@ -31,6 +34,7 @@ static constexpr size_t INPUT_HISTORY_SIZE = 127;
 FilteredTextEdit::FilteredTextEdit(QWidget *parent)
   : QTextEdit{parent}
   , history_index_{0}
+  , isImage_{false}
 {
         connect(document()->documentLayout(),
                 &QAbstractTextDocumentLayout::documentSizeChanged,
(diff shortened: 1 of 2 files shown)


src/TextInputWidget.cc
```diff
@@ -93,12 +97,50 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
                 }
                 break;
         }
+        case Qt::Key_V: {
+                if (event->modifiers() == Qt::ControlModifier) {
```
Ralith: Hardcoding this is incorrect. You cannot make assumptions about what keybind any given system uses to express "paste." Try using `QKeySequence::Copy` instead.

src/TextInputWidget.cc
```diff
@@ -93,12 +97,50 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+}
+
+void
+FilteredTextEdit::insertFromMimeData(const QMimeData *source)
+{
+        if (source->hasImage()) {
+                QImage img = qvariant_cast<QImage>(source->imageData());
+                pastedImagePath_ = QDir::tempPath() + '/' + "nheko_pasted_img.png";
+
+                // Save image into temporary path to be loaded later.
```
Ralith: There's no reason for this to hit disk, and doing so introduces a number of bugs. Leave the image in memory.

christarazi: The reason I write to disk is because `MatrixClient::uploadImage()` takes in a `QString` that represents a file path to an image. This means I would have to overload `uploadImage` or create a new function which accepts an image from memory. If you think that is worth doing, I will implement it.

Ralith: You should probably refactor `uploadImage` to accept data from arbitrary sources, e.g. by taking a [`QIODevice`](https://doc.qt.io/qt-5/qiodevice.html). This change is necessary for a number of reasons; most immediately, this introduces system-global state, which makes it unsafe to have multiple instances of the class even in entirely different processes.

src/TextInputWidget.cc
Ralith: On further review, given that Qt handles MIME clipboards in the first place, it's actually not obvious to me that handling pasting by hand like this is even necessary at all. If it truly is, some comments explaining why Qt's default behavior is inadequate would be nice.

christarazi: I'm unsure of what you mean. If you mean, that Qt has built in support out-of-the-box for pasting images into a QTextEdit, then I have not seen any document which suggests that. The fact that `QTextEdit::insertFromMimeData` and `QTextEdit::canInsertFromMimeData` exists suggests that it does not have built in support. However, this is just an educated guess.

src/TextInputWidget.cc
christarazi: I'll create a new pull request with the requested changes.

src/TextInputWidget.cc
Ralith: No need to guess. What happens if you disable this hardcoding but retain the overrides to those methods?

Ralith:
Another concern is the handling of pastes that include both image(s) and text concatenated in a single piece of MIME data, along with a possible alternative plaintext representation.

It's worth note that Matrix *can* support this type of mixed-content message by placing the alternative plaintext in `body` and using `formatted_body` with `img` tags that reference `mxc` URIs to encode the mixed form, although no existing client generates such messages automatically. I believe Riot, at least, will nonetheless render such messages as expected.

src/TextInputWidget.cc
christarazi: Good call. I removed the explicit call to `insertFromMimeData()` (the entire case in the switch statement) and it still works.

christarazi:
> Another concern is the handling of pastes that include both image(s) and text concatenated in a single piece of MIME data, along with a possible alternative plaintext representation.
> 
> It's worth note that Matrix can support this type of mixed-content message by placing the alternative plaintext in body and using formatted_body with img tags that reference mxc URIs to encode the mixed form, although no existing client generates such messages automatically. I believe Riot, at least, will nonetheless render such messages as expected.

@Ralith On my local branch, I made the change to allow both images and text in the same message. I did this by checking if `isImage_` is true, and if so, then removing the encoded image ([encoded as U+FFFC](https://code.woboq.org/qt5/qtbase/src/gui/text/qtextdocument.cpp.html#1173)) from the text, and allowing the existing code to take care of the rest.

mujx:
How exactly I can use this feature? If I copy-paste an image into the text field I just get the file path of that image and nothing happens when I hit enter.

src/TextInputWidget.cc
christarazi: Fixed in `[41b887b](https://github.com/mujx/nheko/pull/138/commits/41b887b0554f36f527e2478d2bc205a0ae8fb232)`

christarazi:
> How exactly I can use this feature? If I copy-paste an image into the text field I just get the file path of that image and nothing happens when I hit enter.

Didn't realize that X11 treats copying files differently in regards to MIME. I was testing my implementation by right-clicking an image in my browser and hitting "Copy image". This is now fixed in [
4e8034c`](https://github.com/mujx/nheko/pull/138/commits/4e8034c8dda11f116b06b59f8ff256a0fd38ddfe)

christarazi:
Now that I think of it, we might as well handle all files that are pasted in the text box. The commit in [`4e8034c`](https://github.com/mujx/nheko/pull/138/commits/4e8034c8dda11f116b06b59f8ff256a0fd38ddfe) only handles pasting images, but it should be a trivial change to allow pasting of any file.

mujx:
This also doesn't work on macOS. Instead of the actual image a generic image placeholder is sent.

Another issue is that the image is rendered in the text field and only a thin slice of it is visible which makes it a bad user experience. It would be better to show a confirmation popup to send the image or file (like Telegram does).

I'm against mixing images and text in the same messages cause it would be quite difficult to implement and it doesn't have any real benefit in my opinion.

christarazi:
Interesting, seems like handling MIME data was more convoluted than I thought. I will look into it. 

You can scroll using the mouse wheel to see the image, but I agree, this can be improved. 

Alright, I can revert that commit if that's what you'd prefer. 

On November 28, 2017 4:26:25 AM PST, mujx <notifications@github.com> wrote:
>This also doesn't work on macOS. Instead of the actual image a generic
>image placeholder is sent.
>
>Another issue is that the image is rendered in the text field and only
>a thin slice of it is visible which makes it a bad user experience. It
>would be better to show a confirmation popup to send the image or file
>(like Telegram does).
>
>I'm against mixing images and text in the same messages cause it would
>be quite difficult to implement and it doesn't have any real benefit in
>my opinion.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/mujx/nheko/pull/138#issuecomment-347507731

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

christarazi:
Do you know of a method to test macOS binaries on Linux? I do not own a Mac. 

On November 28, 2017 4:26:25 AM PST, mujx <notifications@github.com> wrote:
>This also doesn't work on macOS. Instead of the actual image a generic
>image placeholder is sent.
>
>Another issue is that the image is rendered in the text field and only
>a thin slice of it is visible which makes it a bad user experience. It
>would be better to show a confirmation popup to send the image or file
>(like Telegram does).
>
>I'm against mixing images and text in the same messages cause it would
>be quite difficult to implement and it doesn't have any real benefit in
>my opinion.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/mujx/nheko/pull/138#issuecomment-347507731

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mujx:
@christarazi I could find a fix for mac later, it's not super critical. The important part is to have that popup when a pasted file or image is detected and possibly allow the user to edit the name of the file before sending it, so nothing would be pasted into the text field.

src/TextInputWidget.cc
```diff
@@ -229,6 +280,9 @@ TextInputWidget::TextInputWidget(QWidget *parent)
         connect(sendFileBtn_, SIGNAL(clicked()), this, SLOT(openFileSelection()));
         connect(input_, &FilteredTextEdit::message, this, &TextInputWidget::sendTextMessage);
         connect(input_, &FilteredTextEdit::command, this, &TextInputWidget::command);
+        connect(input_, &FilteredTextEdit::image, this, [=](QString name) {
+                emit uploadImage(name);
+        });
```
Ralith: I believe this can be written `connect(input_, &FilteredTextEdit::image, this, uploadImage)`

christarazi:
I've created a new PR (#180) which adds a preview dialog for pasting images and also modifies the `MatrixClient::upload*()` members to take in a `QIODevice` instead of a `QString`.

PR #150, opened 2017-12-06
Update clang and Ubuntu 14.04 instructions

alphapapa:
On Ubuntu 14.04, Clang 3.6 is the minimum version, and seems to be the only way to build nheko out-of-the-box.  Using GCC on this version of Ubuntu requires installing newer GCC packages from third-party PPAs, which is quite messy and not generally recommended.


mujx:
Thanks a lot!

PR #180, opened 2017-12-24
[Updated] Implements #132: Pasting images into room

christarazi:
This is an update to PR #138

### DIFF
--- a/include/MatrixClient.h
+++ b/include/MatrixClient.h
@@ -52,9 +52,15 @@ public:
         void downloadImage(const QString &event_id, const QUrl &url);
         void downloadFile(const QString &event_id, const QUrl &url);
         void messages(const QString &room_id, const QString &from_token, int limit = 30) noexcept;
-        void uploadImage(const QString &roomid, QSharedPointer<QIODevice> iodev);
-        void uploadFile(const QString &roomid, QSharedPointer<QIODevice> iodev);
-        void uploadAudio(const QString &roomid, QSharedPointer<QIODevice> iodev);
+        void uploadImage(const QString &roomid,
+                         const QSharedPointer<QIODevice> data,
+                         const QString &filename);
+        void uploadFile(const QString &roomid,
+                        const QSharedPointer<QIODevice> data,
+                        const QString &filename);
+        void uploadAudio(const QString &roomid,
+                         const QSharedPointer<QIODevice> data,
+                         const QString &filename);
         void joinRoom(const QString &roomIdOrAlias);
         void leaveRoom(const QString &roomId);
         void sendTypingNotification(const QString &roomid, int timeoutInMillis = 20000);
@@ -94,9 +100,12 @@ signals:
                              const QString &homeserver,
                              const QString &token);
         void versionSuccess();
-        void imageUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
-        void fileUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
-        void audioUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
+        void imageUploaded(const QString &roomid,
+                           const QSharedPointer<QIODevice> data,
+                           const QString &filename,
+                           const QString &url);
+        void fileUploaded(const QString &roomid, const QString &filename, const QString &url);
+        void audioUploaded(const QString &roomid, const QString &filename, const QString &url);
 
         void roomAvatarRetrieved(const QString &roomid,
                                  const QPixmap &img,
--- a/include/MatrixClient.h
+++ b/include/MatrixClient.h
@@ -52,9 +52,9 @@ public:
         void downloadImage(const QString &event_id, const QUrl &url);
         void downloadFile(const QString &event_id, const QUrl &url);
         void messages(const QString &room_id, const QString &from_token, int limit = 30) noexcept;
-        void uploadImage(const QString &roomid, const QString &filename);
-        void uploadFile(const QString &roomid, const QString &filename);
-        void uploadAudio(const QString &roomid, const QString &filename);
+        void uploadImage(const QString &roomid, QSharedPointer<QIODevice> iodev);
+        void uploadFile(const QString &roomid, QSharedPointer<QIODevice> iodev);
+        void uploadAudio(const QString &roomid, QSharedPointer<QIODevice> iodev);
         void joinRoom(const QString &roomIdOrAlias);
         void leaveRoom(const QString &roomId);
         void sendTypingNotification(const QString &roomid, int timeoutInMillis = 20000);
@@ -94,9 +94,9 @@ signals:
                              const QString &homeserver,
                              const QString &token);
         void versionSuccess();
-        void imageUploaded(const QString &roomid, const QString &filename, const QString &url);
-        void fileUploaded(const QString &roomid, const QString &filename, const QString &url);
-        void audioUploaded(const QString &roomid, const QString &filename, const QString &url);
+        void imageUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
+        void fileUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
+        void audioUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
 
         void roomAvatarRetrieved(const QString &roomid,
                                  const QPixmap &img,
@@ -123,7 +123,7 @@ signals:
         void roomCreationFailed(const QString &msg);
 
 private:
-        QNetworkReply *makeUploadRequest(const QString &filename);
+        QNetworkReply *makeUploadRequest(QSharedPointer<QIODevice> iodev);
 
         // Client API prefix.
         QString clientApiUrl_;
--- a/include/dialogs/PreviewImageOverlay.h
+++ b/include/dialogs/PreviewImageOverlay.h
@@ -17,12 +17,14 @@
 
 #pragma once
 
+#include <QLabel>
+#include <QLineEdit>
 #include <QPixmap>
 #include <QWidget>
 
-class QLabel;
-class QLineEdit;
-class QPushButton;
+#include "FlatButton.h"
+
+class QMimeData;
 
 namespace dialogs {
 
@@ -30,22 +32,26 @@ class PreviewImageOverlay : public QWidget
 {
         Q_OBJECT
 public:
-        PreviewImageOverlay(QPixmap image, QWidget *parent = nullptr);
-        PreviewImageOverlay(const QString &path, QWidget *parent = nullptr);
+        PreviewImageOverlay(QWidget *parent = nullptr);
+
+        void setImageAndCreate(const QByteArray &data, const QString &type);
+        void setImageAndCreate(const QString &path);
 
 signals:
-        void confirmImageUpload(const QPixmap &img, const QString &img_name);
+        void confirmImageUpload(const QByteArray &data, const QString &img_name);
 
 private:
         void init();
 
         QPixmap image_;
+        QByteArray imageData_;
+        QString imagePath_;
 
-        QLabel *titleLabel_;
-        QLabel *imageLabel_;
-        QLineEdit *imageName_;
+        QLabel titleLabel_;
+        QLabel imageLabel_;
+        QLineEdit imageName_;
 
-        QPushButton *upload_;
-        QPushButton *cancel_;
+        FlatButton upload_;
+        FlatButton cancel_;
 };
 } // dialogs
--- a/include/dialogs/PreviewImageOverlay.h
+++ b/include/dialogs/PreviewImageOverlay.h
@@ -1,0 +1,51 @@
+/*
+ * nheko Copyright (C) 2017  Konstantinos Sideris <siderisk@auth.gr>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <QPixmap>
+#include <QWidget>
+
+class QLabel;
+class QLineEdit;
+class QPushButton;
+
+namespace dialogs {
+
+class PreviewImageOverlay : public QWidget
+{
+        Q_OBJECT
+public:
+        PreviewImageOverlay(QPixmap image, QWidget *parent = nullptr);
+        PreviewImageOverlay(const QString &path, QWidget *parent = nullptr);
+
+signals:
+        void confirmImageUpload(const QPixmap &img, const QString &img_name);
+
+private:
+        void init();
+
+        QPixmap image_;
+
+        QLabel *titleLabel_;
+        QLabel *imageLabel_;
+        QLineEdit *imageName_;
+
+        QPushButton *upload_;
+        QPushButton *cancel_;
+};
+} // dialogs
--- a/include/dialogs/PreviewImageOverlay.h
+++ b/include/dialogs/PreviewImageOverlay.h
@@ -34,11 +34,11 @@ class PreviewImageOverlay : public QWidget
 public:
         PreviewImageOverlay(QWidget *parent = nullptr);
 
-        void setImageAndCreate(const QByteArray &data, const QString &type);
+        void setImageAndCreate(const QByteArray data, const QString &type);
         void setImageAndCreate(const QString &path);
 
 signals:
-        void confirmImageUpload(const QByteArray &data, const QString &img_name);
+        void confirmImageUpload(const QByteArray data, const QString &img_name);
 
 private:
         void init();
--- a/include/timeline/TimelineView.h
+++ b/include/timeline/TimelineView.h
@@ -43,20 +43,20 @@ struct PendingMessage
         mtx::events::MessageType ty;
         int txn_id;
         QString body;
-        QString filename;
+        QSharedPointer<QFile> file;
         QString event_id;
         TimelineItem *widget;
 
         PendingMessage(mtx::events::MessageType ty,
                        int txn_id,
                        QString body,
-                       QString filename,
+                       QSharedPointer<QFile> file,
                        QString event_id,
                        TimelineItem *widget)
           : ty(ty)
           , txn_id(txn_id)
           , body(body)
-          , filename(filename)
+          , file(file)
           , event_id(event_id)
           , widget(widget)
         {}
@@ -87,7 +87,7 @@ public:
         void addUserMessage(mtx::events::MessageType ty, const QString &msg);
 
         template<class Widget, mtx::events::MessageType MsgType>
-        void addUserMessage(const QString &url, const QString &filename);
+        void addUserMessage(const QString &url, QSharedPointer<QFile> file);
         void updatePendingMessage(int txn_id, QString event_id);
         void scrollDown();
         void addDateSeparator(QDateTime datetime, int position);
@@ -216,11 +216,11 @@ private:
 
 template<class Widget, mtx::events::MessageType MsgType>
 void
-TimelineView::addUserMessage(const QString &url, const QString &filename)
+TimelineView::addUserMessage(const QString &url, QSharedPointer<QFile> file)
 {
         auto with_sender = lastSender_ != local_user_;
 
-        auto widget = new Widget(client_, url, filename, this);
+        auto widget = new Widget(client_, url, file, this);
 
         TimelineItem *view_item =
           new TimelineItem(widget, local_user_, with_sender, scroll_widget_);
@@ -234,7 +234,7 @@ TimelineView::addUserMessage(const QString &url, const QString &filename)
 
         int txn_id = client_->incrementTransactionId();
 
-        PendingMessage message(MsgType, txn_id, url, filename, "", view_item);
+        PendingMessage message(MsgType, txn_id, url, file, "", view_item);
         handleNewUserMessage(message);
 }
 
--- a/include/timeline/TimelineView.h
+++ b/include/timeline/TimelineView.h
@@ -50,13 +50,13 @@ struct PendingMessage
         PendingMessage(mtx::events::MessageType ty,
                        int txn_id,
                        QString body,
-                       const QString &fn,
+                       QString filename,
                        QString event_id,
                        TimelineItem *widget)
           : ty(ty)
           , txn_id(txn_id)
           , body(body)
-          , filename(fn)
+          , filename(filename)
           , event_id(event_id)
           , widget(widget)
         {}
--- a/include/timeline/TimelineView.h
+++ b/include/timeline/TimelineView.h
@@ -59,8 +59,7 @@ struct PendingMessage
           , filename(filename)
           , event_id(event_id)
           , widget(widget)
-        {
-        }
+        {}
 };
 
 // In which place new TimelineItems should be inserted.
--- a/include/timeline/TimelineView.h
+++ b/include/timeline/TimelineView.h
@@ -43,20 +43,20 @@ struct PendingMessage
         mtx::events::MessageType ty;
         int txn_id;
         QString body;
-        QSharedPointer<QFile> file;
+        QString filename;
         QString event_id;
         TimelineItem *widget;
 
         PendingMessage(mtx::events::MessageType ty,
                        int txn_id,
                        QString body,
-                       QSharedPointer<QFile> file,
+                       const QString &fn,
                        QString event_id,
                        TimelineItem *widget)
           : ty(ty)
           , txn_id(txn_id)
           , body(body)
-          , file(file)
+          , filename(fn)
           , event_id(event_id)
           , widget(widget)
         {}
@@ -87,7 +87,9 @@ public:
         void addUserMessage(mtx::events::MessageType ty, const QString &msg);
 
         template<class Widget, mtx::events::MessageType MsgType>
-        void addUserMessage(const QString &url, QSharedPointer<QFile> file);
+        void addUserMessage(const QString &url,
+                            const QSharedPointer<QIODevice> data,
+                            const QString &filename);
         void updatePendingMessage(int txn_id, QString event_id);
         void scrollDown();
         void addDateSeparator(QDateTime datetime, int position);
@@ -216,11 +218,13 @@ private:
 
 template<class Widget, mtx::events::MessageType MsgType>
 void
-TimelineView::addUserMessage(const QString &url, QSharedPointer<QFile> file)
+TimelineView::addUserMessage(const QString &url,
+                             const QSharedPointer<QIODevice> data,
+                             const QString &filename)
 {
         auto with_sender = lastSender_ != local_user_;
 
-        auto widget = new Widget(client_, url, file, this);
+        auto widget = new Widget(client_, url, data, filename, this);
 
         TimelineItem *view_item =
           new TimelineItem(widget, local_user_, with_sender, scroll_widget_);
@@ -234,7 +238,7 @@ TimelineView::addUserMessage(const QString &url, QSharedPointer<QFile> file)
 
         int txn_id = client_->incrementTransactionId();
 
-        PendingMessage message(MsgType, txn_id, url, file, "", view_item);
+        PendingMessage message(MsgType, txn_id, url, filename, "", view_item);
         handleNewUserMessage(message);
 }
 
--- a/include/timeline/TimelineView.h
+++ b/include/timeline/TimelineView.h
@@ -59,7 +59,8 @@ struct PendingMessage
           , filename(filename)
           , event_id(event_id)
           , widget(widget)
-        {}
+        {
+        }
 };
 
 // In which place new TimelineItems should be inserted.
--- a/include/timeline/widgets/FileItem.h
+++ b/include/timeline/widgets/FileItem.h
@@ -42,7 +42,8 @@ public:
 
         FileItem(QSharedPointer<MatrixClient> client,
                  const QString &url,
-                 QSharedPointer<QFile> file,
+                 const QSharedPointer<QIODevice> data,
+                 const QString &filename,
                  QWidget *parent = nullptr);
 
         QSize sizeHint() const override;
--- a/include/timeline/widgets/FileItem.h
+++ b/include/timeline/widgets/FileItem.h
@@ -42,7 +42,7 @@ public:
 
         FileItem(QSharedPointer<MatrixClient> client,
                  const QString &url,
-                 const QString &filename,
+                 QSharedPointer<QFile> file,
                  QWidget *parent = nullptr);
 
         QSize sizeHint() const override;
--- a/resources/styles/nheko-dark.qss
+++ b/resources/styles/nheko-dark.qss
@@ -85,11 +85,8 @@ dialogs--LeaveRoom,
 dialogs--CreateRoom,
 dialogs--InviteUsers,
 dialogs--ReadReceipts,
-dialogs--JoinRoom {
-    background-color: #383c4a;
-    color: #caccd1;
-}
-
+dialogs--JoinRoom,
+dialogs--PreviewImageOverlay,
 QListWidget {
     background-color: #383c4a;
     color: #caccd1;
--- a/resources/styles/nheko-dark.qss
+++ b/resources/styles/nheko-dark.qss
@@ -86,12 +86,12 @@ dialogs--CreateRoom,
 dialogs--InviteUsers,
 dialogs--ReadReceipts,
 dialogs--JoinRoom,
-dialogs--PreviewImageOverlay,
-QListWidget {
+dialogs--PreviewImageOverlay {
     background-color: #383c4a;
     color: #caccd1;
 }
 
+QListWidget,
 WelcomePage,
 LoginPage,
 RegisterPage {
--- a/src/MatrixClient.cc
+++ b/src/MatrixClient.cc
@@ -1028,7 +1028,7 @@ MatrixClient::makeUploadRequest(QSharedPointer<QIODevice> iodev)
         endpoint.setQuery(query);
 
         if (!iodev->open(QIODevice::ReadOnly)) {
-                qDebug() << "Error while reading buffer" << iodev->errorString();
+                qWarning() << "Error while reading device:" << iodev->errorString();
                 return nullptr;
         }
 
@@ -1036,7 +1036,6 @@ MatrixClient::makeUploadRequest(QSharedPointer<QIODevice> iodev)
         QMimeType mime = db.mimeTypeForData(iodev.data());
 
         QNetworkRequest request(QString(endpoint.toEncoded()));
-        request.setHeader(QNetworkRequest::ContentLengthHeader, iodev->size());
         request.setHeader(QNetworkRequest::ContentTypeHeader, mime.name());
 
         auto reply = post(request, iodev.data());
--- a/src/MatrixClient.cc
+++ b/src/MatrixClient.cc
@@ -675,14 +675,14 @@ MatrixClient::messages(const QString &roomid, const QString &from_token, int lim
 }
 
 void
-MatrixClient::uploadImage(const QString &roomid, const QString &filename)
+MatrixClient::uploadImage(const QString &roomid, QSharedPointer<QIODevice> iodev)
 {
-        auto reply = makeUploadRequest(filename);
+        auto reply = makeUploadRequest(iodev);
 
         if (reply == nullptr)
                 return;
 
-        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, filename]() {
+        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, iodev]() {
                 reply->deleteLater();
 
                 int status = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
@@ -711,16 +711,16 @@ MatrixClient::uploadImage(const QString &roomid, const QString &filename)
                         return;
                 }
 
-                emit imageUploaded(roomid, filename, object.value("content_uri").toString());
+                emit imageUploaded(roomid, iodev.dynamicCast<QFile>(), object.value("content_uri").toString());
         });
 }
 
 void
-MatrixClient::uploadFile(const QString &roomid, const QString &filename)
+MatrixClient::uploadFile(const QString &roomid, QSharedPointer<QIODevice> iodev)
 {
-        auto reply = makeUploadRequest(filename);
+        auto reply = makeUploadRequest(iodev);
 
-        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, filename]() {
+        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, iodev]() {
                 reply->deleteLater();
 
                 int status = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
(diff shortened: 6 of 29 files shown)


mujx:
Thanks a lot for your work! The preview modal is really good 👍 

The only issue I spotted was with the file naming. On the dialog the default name should be the local filename. Also when I send the image it appears with the temporary filename and not with the one I gave. You can pass around the full local path, I don't think there is a need to create a temporary file.

include/MatrixClient.h
```diff
@@ -94,9 +94,9 @@ public slots:
                              const QString &homeserver,
                              const QString &token);
         void versionSuccess();
-        void imageUploaded(const QString &roomid, const QString &filename, const QString &url);
-        void fileUploaded(const QString &roomid, const QString &filename, const QString &url);
-        void audioUploaded(const QString &roomid, const QString &filename, const QString &url);
+        void imageUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
+        void fileUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
+        void audioUploaded(const QString &roomid, QSharedPointer<QFile> file, const QString &url);
```
Ralith: Why are these `QFile`? The whole point of the change to use `QIODevice` is that the uploaded data is not necessarily a file. As the filename (and only the filename) is needed for UI purposes, this should remain a `QString`.

include/MatrixClient.h
```diff
@@ -52,9 +52,9 @@ class MatrixClient : public QNetworkAccessManager
         void downloadImage(const QString &event_id, const QUrl &url);
         void downloadFile(const QString &event_id, const QUrl &url);
         void messages(const QString &room_id, const QString &from_token, int limit = 30) noexcept;
-        void uploadImage(const QString &roomid, const QString &filename);
-        void uploadFile(const QString &roomid, const QString &filename);
-        void uploadAudio(const QString &roomid, const QString &filename);
+        void uploadImage(const QString &roomid, QSharedPointer<QIODevice> iodev);
+        void uploadFile(const QString &roomid, QSharedPointer<QIODevice> iodev);
+        void uploadAudio(const QString &roomid, QSharedPointer<QIODevice> iodev);
```
Ralith: Why do these need to be shared pointers?

src/TextInputWidget.cc
```diff
@@ -170,6 +208,30 @@ FilteredTextEdit::textChanged()
...
+        if (!img.save(file->fileName(), "PNG")) {
+                qDebug() << "Failed to save image data to:" << file->fileName();
+                previewDialog_->close();
+                return;
+        }
+        file->close();
+
+        emit image(file);
+
+        previewDialog_->close();
```
Ralith: Since you close the preview dialog no matter what, wouldn't it be simpler to do so in one place, in the dialog's implementation at the same time you emit `confirmImageUpload`? That way it's impossible to forget, and much less likely to be broken by future changes.

src/TextInputWidget.cc
```diff
@@ -101,6 +105,39 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+        if (source->hasImage()) {
+                QPixmap const img = qvariant_cast<QPixmap>(source->imageData());
+                previewDialog_    = new dialogs::PreviewImageOverlay{img, this};
+                connect(previewDialog_,
+                        &dialogs::PreviewImageOverlay::confirmImageUpload,
+                        this,
+                        &FilteredTextEdit::receiveImage);
+                previewDialog_->show();
+        } else if (source->hasFormat("x-special/gnome-copied-files") &&
+                   QImageReader{source->text()}.canRead()) {
```
Ralith: What if a user pastes in non-image files?

src/TextInputWidget.cc
```diff
@@ -170,6 +208,30 @@ FilteredTextEdit::textChanged()
...
+        // Save image into temporary path to be loaded later.
+        QString img_path = QDir::tempPath() + '/' + img_name;
+        QSharedPointer<QFile> file{new QFile{img_path, this}};
+
+        if (!file->open(QIODevice::WriteOnly)) {
+                qDebug() << "Failed to open temporary image file:" << file->errorString();
+                previewDialog_->close();
+                return;
+        }
+        if (!img.save(file->fileName(), "PNG")) {
```
Ralith: Saving the file to disk remains an unnecessary pessimization, and it's further unclear how you are guaranteeing that the temporary files are deleted when no longer needed.

You can encode images without touching the disk using a [QImageWriter](https://doc.qt.io/qt-5/qimagewriter.html) (for example).

src/TextInputWidget.cc
```diff
@@ -101,6 +105,39 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+        return (source->hasImage() || source->hasText() ||
+                QTextEdit::canInsertFromMimeData(source));
+}
+
+void
+FilteredTextEdit::insertFromMimeData(const QMimeData *source)
+{
+        if (source->hasImage()) {
+                QPixmap const img = qvariant_cast<QPixmap>(source->imageData());
+                previewDialog_    = new dialogs::PreviewImageOverlay{img, this};
```
Ralith: When does the `PreviewImageOverlay` you create here get deleted? Why does it need to be dynamically allocated at all?

src/TextInputWidget.cc
```diff
@@ -101,6 +105,39 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+                connect(previewDialog_,
+                        &dialogs::PreviewImageOverlay::confirmImageUpload,
+                        this,
+                        &FilteredTextEdit::receiveImage);
+                previewDialog_->show();
+        } else if (source->hasFormat("x-special/gnome-copied-files") &&
+                   QImageReader{source->text()}.canRead()) {
+                // Special case for X11 users. See "Notes for X11 Users" in source.
+                // Source: http://doc.qt.io/qt-5/qclipboard.html
+                previewDialog_ = new dialogs::PreviewImageOverlay{source->text(), this};
```
Ralith: See above re: deletion.

src/TextInputWidget.cc
```diff
@@ -101,6 +105,39 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+{
+        return (source->hasImage() || source->hasText() ||
+                QTextEdit::canInsertFromMimeData(source));
+}
+
+void
+FilteredTextEdit::insertFromMimeData(const QMimeData *source)
+{
+        if (source->hasImage()) {
+                QPixmap const img = qvariant_cast<QPixmap>(source->imageData());
```
Ralith: [`QMimeData::imageData`](https://doc.qt.io/qt-5/qmimedata.html#imageData) is documented as returning a `QVariant` containing a `QImage`. Why is a `QPixmap` used here instead?

src/MatrixClient.cc
```diff
@@ -711,16 +711,16 @@ MatrixClient::uploadImage(const QString &roomid, const QString &filename)
                         return;
                 }
 
-                emit imageUploaded(roomid, filename, object.value("content_uri").toString());
+                emit imageUploaded(roomid, iodev.dynamicCast<QFile>(), object.value("content_uri").toString());
```
Ralith: It's unclear whether there's any handling for the case where this cast fails and returns an empty `QSharedPointer`. Regardless, as discussed elsewhere, `QFile` should not be used here.

src/timeline/widgets/ImageItem.cc
```diff
@@ -61,11 +61,11 @@ ImageItem::ImageItem(QSharedPointer<MatrixClient> client,
 
 ImageItem::ImageItem(QSharedPointer<MatrixClient> client,
                      const QString &url,
-                     const QString &filename,
+                     QSharedPointer<QFile> file,
```
Ralith: How will this handle images that aren't written to disk?

include/MatrixClient.h
christarazi: They're shared pointers because notice the lambda in the `MatrixClient::upload*()` member functions for responding to a `QNetworkReply::finished` signal. By the time that lambda is executed (asynchronously), we have lost the scope of the `QIODevice`. If I were to pass by const ref, the reference would be bound to a temporary local, which would cause a segfault somewhere during the execution of the lambda. To resolve this, the best method I could think of was to create a pointer so that it's allocated on the heap. 

I would love to hear if you have a better method.

include/MatrixClient.h
christarazi: Ah this is an oversight on my end. You are correct, they do not need to take a `QFile`. However, it is not clear to me what the `QString` should be, if the data does not come from a file (where the string would obviously be the file name).

Ralith: I'd just invent something descriptive, depending on the actual origin of the data. For example, this PR might use the name "clipboard.png" for UI and upload filename purposes.

src/TextInputWidget.cc
christarazi: The reason it is dynamically allocated is because the dialog is created on a paste event, which is dynamic behaviour by nature. The same approach is used in [`src/SideBarActions.cc`](https://github.com/mujx/nheko/blob/master/src/SideBarActions.cc#L39) and [`src/UserInfoWidget.cc`](https://github.com/mujx/nheko/blob/master/src/UserInfoWidget.cc#L95), to name a few examples.

How it is deleted is an oversight. I have locally changed `previewDialog_` to be a `QSharedPointer` so that it can be de-allocated automatically. I also pass `this` as a parameter to the constructor as to create a parent-child relationship with Qt's object model.

Please let me know if you have a better approach.

include/MatrixClient.h
christarazi: Just to clarify, how would I know if I received a file or a buffer containing data? Would the best method be to try to cast to a `QFile` and use the file name if it exists, and if the cast fails, then use the "clipboard.png" mentioned?

src/TextInputWidget.cc
Ralith: The UI semantics you describe are orthogonal to the actual construction/destruction of the object; you could let `PreviewImageOverlay` be a regular class member, and just show/hide/reset it as necessary. Dynamic allocation can be simpler, but requires care to avoid introducing leaks. Here, using Qt's parent-child relationship does not avoid a leak because many pastes may occur in the lifetime of a `TextInputWidget`.

Switching to a `QSharedPointer` fixes the problem, but may be confusing, as it's presumably not intended that ownership of the value be shared. Instead, I'd suggest using `std::unique_ptr`, and furthermore replacing the call to `new` with `std::make_unique`; calling `new` explicitly should generally be avoided, as it's prone to this type of mistake.

include/MatrixClient.h
Ralith: You should not need to know what specific `QIODevice` generic code receives; doing so defeats the purpose of having generic code. For example, the approach you outline will break down once there are non-clipboard sources of in-memory data (e.g. imagery captured directly from a camera). Justifiable uses of dynamic casting in C++ are extremely rare.

Instead, whenever a filename will eventually be needed alongside a `QIODevice`, you should pass a filename along from the very beginning. For example, the signature of `FilteredTextEdit::image` could be changed to `void FilteredTextEdit::image(QSharedPointer<QIODevice> data, const QString &filename)`

src/TextInputWidget.cc
christarazi: As of now, pasting non-image files will trigger the else clause and therefore, Qt's default behaviour will take over. I think adding support for pasting non-image files (as well as audio, video, etc) is fairly simple, and probably out of the scope of this PR. It is definitely something I can look into in the future.

src/timeline/widgets/ImageItem.cc
christarazi: I am unsure how to approach this yet as I did not foresee this as a problem. 

To properly load the image, we also need access to the image data. It's not a difficult change, but requires numerous functions to be modified to be able to pass the image data along with the filename. I would like to hear your thoughts on this one.

Ralith: Can you give a few examples of functions that would need to be modified? An image in the process of being uploaded shouldn't need to be accessed from very many places, and I don't imagine there are many places other than the timeline or the upload procedure itself where a filename is pertinent.

christarazi: The following is the calling order which would need to be modified to pass the image data:

1) MatrixClient::imageUploaded
2) TimelineViewManager::queueImageMessage
3) TimelineView::addUserMessage
4) ImageItem constructor

Ralith: That doesn't seem particularly egregious to me. If the increase in number of parameters bothers you, you could always create a `struct Image { QImage data; QString name; };` or similar to bundle them up for convenience.

christarazi:
I've added a new commit which hopefully address all the issues pointed out.

src/TextInputWidget.cc
```diff
@@ -101,6 +111,51 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+
+void
+FilteredTextEdit::insertFromMimeData(const QMimeData *source)
+{
+        if (source->hasImage()) {
+                const QImage image = qvariant_cast<QImage>(source->imageData());
+
+                const auto formats = source->formats();
+                const auto idx     = formats.indexOf(
+                  QRegularExpression{"image/\\w+", QRegularExpression::CaseInsensitiveOption});
```
Ralith: Is there a reason to be more selective than `image/.*`?

src/TextInputWidget.cc
```diff
@@ -101,6 +111,51 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
...
+{
+        return (source->hasImage() || source->hasText() ||
+                QTextEdit::canInsertFromMimeData(source));
+}
+
+void
+FilteredTextEdit::insertFromMimeData(const QMimeData *source)
+{
+        if (source->hasImage()) {
+                const QImage image = qvariant_cast<QImage>(source->imageData());
```
Ralith: On further investigation, it's apparent to me that we can and therefore must use [`QMimeData::data`](https://doc.qt.io/qt-5/qmimedata.html#data) to avoid re-encoding the image at all, rather than decoding the image here and re-encoding it below. This will avoid silent quality loss, improve robustness, and reduce CPU and memory use. The mime type obtained from `QMimeData` should also be propagated alongside the data so it can be included in the matrix event verbatim, removing any need to guess it.

I apologize for not identifying this option earlier; I'm not super familiar with how rich copy/paste works.

include/timeline/TimelineView.h
```diff
@@ -51,13 +51,13 @@ struct PendingMessage
         PendingMessage(mtx::events::MessageType ty,
                        int txn_id,
                        QString body,
-                       QString filename,
+                       const QString &fn,
```
Ralith: Why was this argument name changed?

resources/styles/nheko-dark.qss
```diff
@@ -84,12 +84,8 @@ dialogs--Logout,
 dialogs--LeaveRoom,
 dialogs--CreateRoom,
 dialogs--InviteUsers,
-dialogs--JoinRoom {
-    background-color: #383c4a;
-    color: #caccd1;
-}
-
-QListWidget {
```
Ralith: Why was this line deleted?

include/timeline/widgets/FileItem.h
```diff
@@ -42,6 +42,7 @@ class FileItem : public QWidget
 
         FileItem(QSharedPointer<MatrixClient> client,
                  const QString &url,
+                 const QSharedPointer<QIODevice> data,
```
Ralith: Why is this needed? Unlike with image/audio/video, I'm not sure we can/should do much in the client with the data of an arbitrary opaque file.

src/MatrixClient.cc
```diff
@@ -1021,20 +1027,19 @@ MatrixClient::makeUploadRequest(const QString &filename)
...
-        QMimeType mime = db.mimeTypeForFile(filename, QMimeDatabase::MatchContent);
+        QMimeType mime = db.mimeTypeForData(iodev.data());
 
         QNetworkRequest request(QString(endpoint.toEncoded()));
-        request.setHeader(QNetworkRequest::ContentLengthHeader, file.size());
+        request.setHeader(QNetworkRequest::ContentLengthHeader, iodev->size());
         request.setHeader(QNetworkRequest::ContentTypeHeader, mime.name());
 
-        auto reply = post(request, file.readAll());
+        auto reply = post(request, iodev->readAll());
```
Ralith: It's inappropriate to read the `QIODevice` fully into memory before transmitting. This wastes time and memory. Instead, it should be streamed directly. Note that [`QNetworkManager::post`](https://doc.qt.io/qt-5/qnetworkaccessmanager.html#post) supports this directly.

src/TextInputWidget.cc
```diff
@@ -170,6 +226,13 @@ FilteredTextEdit::textChanged()
         working_history_[history_index_] = toPlainText();
 }
 
+void
+FilteredTextEdit::receiveImage(const QByteArray &img, const QString &img_name)
+{
+        QSharedPointer<QBuffer> buffer{new QBuffer{const_cast<QByteArray *>(&img), this}};
```
Ralith: In addition to being const-unsafe (it is almost never correct to use `const_cast`), this exposes a pointer to the `imageData_` member of `PreviewImageOverlay`. What happens if the user pastes in another image while the emitted `QBuffer` is still in use somewhere?

Instead, create a `QBuffer` that owns its own storage using its default constructor, and initialize it by calling [`QBuffer::setData`](https://doc.qt.io/qt-5/qbuffer.html#setData).

include/dialogs/PreviewImageOverlay.h
```diff
@@ -0,0 +1,57 @@
...
+{
+        Q_OBJECT
+public:
+        PreviewImageOverlay(QWidget *parent = nullptr);
+
+        void setImageAndCreate(const QByteArray &data, const QString &type);
+        void setImageAndCreate(const QString &path);
+
+signals:
+        void confirmImageUpload(const QByteArray &data, const QString &img_name);
```
Ralith: This signal should emit a `QByteArray` by value for reasons discussed above.

include/dialogs/PreviewImageOverlay.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace dialogs {
+
+class PreviewImageOverlay : public QWidget
+{
+        Q_OBJECT
+public:
+        PreviewImageOverlay(QWidget *parent = nullptr);
+
+        void setImageAndCreate(const QByteArray &data, const QString &type);
```
Ralith: To avoid inadvertent sharing and possible UB, this should take a `QByteArray` by value. [`QByteArray` is internally ref-counted](https://doc.qt.io/qt-5/qbytearray.html#QByteArray-4) so doing so is cheap.

alphapapa:
@Ralith I'm just an interested user, but thanks for your very thorough reviews of this patch.  I'm glad that nheko is getting such high-quality code!  (I wish I had a reviewer like you for my own projects!)

include/timeline/widgets/FileItem.h
christarazi: This is needed because [`TimelineView::addUserMessage`](https://github.com/mujx/nheko/pull/180/commits/5ccc484cea0f239c39a7e0f634b463048e059083#diff-f4908a06f6a57ca0c9989bd84b42b7c3R206) takes in a generic type `Widget`, which happens to be one of the four `FileItem`, `ImageItem`, `AudioItem`, and `VideoItem`. Therefore, we need those types to have the same interface (parameters to the constructor).

src/TextInputWidget.cc
christarazi: Forgive me if I'm missing something here, but using `QMimeData::data` won't prevent re-encoding, because we must pass a `QString` representing the MIME type, which is then used to create (encode the raw pixels) the `QByteArray`. I also am not too familiar with copy/paste media, but I've noticed that when copying an image (not from a file manager), it will copy the raw pixel bytes, understandably, as there's no notion of a file. I only came to this realization when creating this PR. 

Therefore, re-encoding cannot be avoided. However, your suggestion has reduced the amount of code necessary and made it much cleaner.

christarazi:
@Ralith I would like to echo @alphapapa 's comment. I must admit your attitude and determination for quality code is quite inspiring and motivating, and exhibits one of the many qualities that I love about free and open source software and their respective communities. So thank you for your hard work and I wish everyone a happy new year. :)

include/timeline/widgets/FileItem.h
Ralith: I'm not particularly happy about having never-used arguments making things confusing on both ends for the sake of convenience to a template. That said, refactoring `addUserMessage` is probably out of scope for this PR.

src/TextInputWidget.cc
Ralith: > we must pass a `QString` representing the MIME type, which is then used to create (encode the raw pixels) the `QByteArray`

My understanding is that `QMimeData::data` performs no transcoding. The supplied `mimeType` is used to select *which* of multiple pieces of attached data you want (from the alternatives exposed by `formats`). This can be observed in the [`QMimeDataPrivate::retrieveTypedData`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qmimedata.cpp#n108) implementation's  handling of the `type == QVariant::ByteArray` case, which for image data falls through ultimately to [`QMimeDataPrivate::getData`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qmimedata.cpp#n96), which is a simple linear search through `dataList`.

> I also am not too familiar with copy/pasting media, but I've noticed that when copying an image (not from a file manager), it will copy the raw pixel bytes, understandably, as there's no notion of a file.

What was your methodology here? If we're genuinely receiving `image/*` mime types, as the current PR suggests, I would expect the data to be in the advertised format. In fact, reviewing the implementation, it appears to me that [`QMimeData::imageData` returns nonempty results *only* when the mime type is `application/x-qt-image`](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qmimedata.cpp#n478). PNG data certainly doesn't have to be physically stored in a file to be valid.

*time passes*

Right clicking on a jpg image in Chrome, selecting "Copy image," and inspecting the clipboard state with `xclip -o -selection clipboard -t TARGETS` reveals that the clipboard contains formats, `image/png` and `text/html`. Dumping the former yields a valid png image, and the latter yields a brief HTML document containing an image tag. I'm a bit perplexed by Chrome's apparent decision to transcode the image, but it's apparent that we do not need to.

> re-encoding cannot be avoided

If uncompressed data is actually the norm, we should take care to encode it as png (or perhaps something more modern, like webp?) for sanity's sake, but per the above testing this does not appear to be the case.

Ralith:
Thank you both for saying so! I'm glad my comments have been helpful.

alphapapa:
@christarazi Forgive me for neglecting to thank *you* for making this patch as well!  :)

src/TextInputWidget.cc
christarazi: Huh, seems I have drawn an incorrect conclusion because of a misinterpretation of the `QImage::bits` method. These are the steps that I took which led me to make my initial claim:

```cpp
QImage img = qvariant_cast<QImage>(source->imageData());
QByteArray arr = QByteArray{img.bits(), img.byteCount()};
```
For some reason, I thought that this would essentially convert the `QImage` into a byte array. Instead, it gives the actual raw pixel bytes, which led me to incorrectly believe that "Copy image" copies raw bytes even after saving the `QImage` as PNG.

Alright well, I apologize for that, my knowledge of the clipboard is clearly lacking. I was not even aware of that particular capability of `xclip`, that's pretty nifty.

src/TextInputWidget.cc
christarazi: Not necessarily. I just find it more reassuring than one that gives `image/` as a potential match. I do not mind changing it if you prefer.

resources/styles/nheko-dark.qss
christarazi: Oops that was accidental. It probably slipped in when I was doing a rebase onto upstream from my previous PR. It has been fixed.

src/MatrixClient.cc
christarazi: I was not aware of that particular overload. Fixed.

src/TextInputWidget.cc
Ralith: I think `QImage` probably *is* uncompressed data, since it supports editing. I'm still a bit confused about whether `imageData` is ever nonempty when pasting e.g. `image/png`; it seems like it can't be, but I assume you've the original code and it does something at least some of the time. Perhaps Qt automatically decodes incoming pastes for convenience of display.

One potential subtlety to bear in mind is the case where many image formats available. For example, here's the list firefox supplies:
```
image/png
image/bmp
image/x-bmp
image/x-MS-bmp
image/jpeg
image/tiff
image/x-icon
image/x-ico
image/x-win-bitmap
image/vnd.microsoft.icon
application/ico
image/ico
image/icon
text/ico
```
Given that the data supplied by both firefox and chrome seems to be re-encoded (the input was a jpeg image, and the image/jpeg content was 6 times smaller than the input) IMO a lossless format should be preferred. I'm happy to work under the hypothesis that everyone's helpful and orders paste data in descending quality--i.e. that it is correct to pick the first `image` content if any exists--but if this is determined to often be false, we may in the future want to add some heuristics to try to make a good selection. Might be worth leaving a note to ourselves in a comment.

src/TextInputWidget.cc
Ralith: Mainly I'm leery of inadvertently filtering out future valid image types. It's not immediately obvious to me that Qt's interpretation of `\w` is a superset of valid media type characters. I wouldn't object to `image/.+` either, but if something's generating clipboard contents with garbage mime-types anything we do is going to be best-effort anyway.

src/dialogs/PreviewImageOverlay.cc
```diff
@@ -0,0 +1,121 @@
...
+        imageLabel_.setScaledContents(false);
+
+        raise();
+}
+
+void
+PreviewImageOverlay::setImageAndCreate(const QByteArray &data, const QString &type)
+{
+        imageData_ = data;
+        image_.loadFromData(imageData_);
```
Ralith: It would be nice to have a fallback (perhaps a QLabel with some explanatory text) for when this fails, i.e. when the user pasted in an image of a type that Qt doesn't support decoding, or that `QImage` can't accurately guess the format of. In those cases we should still allow uploading, but the current behavior might produce a confusing UI.

src/timeline/widgets/ImageItem.cc
```diff
@@ -83,7 +84,14 @@ ImageItem::ImageItem(QSharedPointer<MatrixClient> client,
         url_                 = QString("%1/_matrix/media/r0/download/%2")
                  .arg(client_.data()->getHomeServer().toString(), media_params);
 
-        setImage(QPixmap(filename));
+        if (!data.isNull()) {
+                data->reset();
```
Ralith: The return code of this shouldn't be ignored unless you can prove that it's guaranteed to succeed.

src/MatrixClient.cc
```diff
@@ -1021,20 +1027,19 @@ MatrixClient::makeUploadRequest(const QString &filename)
         endpoint.setPath(mediaApiUrl_ + "/upload");
         endpoint.setQuery(query);
 
-        QFile file(filename);
-        if (!file.open(QIODevice::ReadWrite)) {
-                qDebug() << "Error while reading" << filename;
+        if (!iodev->open(QIODevice::ReadWrite)) {
+                qDebug() << "Error while reading buffer" << iodev.data();
```
Ralith: Logged errors should contain [the specific error message](https://doc.qt.io/qt-5/qiodevice.html#errorString).

src/timeline/widgets/ImageItem.cc
```diff
@@ -61,11 +61,12 @@ ImageItem::ImageItem(QSharedPointer<MatrixClient> client,
 
 ImageItem::ImageItem(QSharedPointer<MatrixClient> client,
                      const QString &url,
+                     const QSharedPointer<QIODevice> data,
                      const QString &filename,
                      QWidget *parent)
   : QWidget(parent)
   , url_{url}
-  , text_{QFileInfo(filename).fileName()}
+  , text_{QFileInfo{filename}.fileName()}
```
Ralith: `filename` is no longer a path on disk, so this won't work.

src/MatrixClient.cc
christarazi: I am having trouble with this one actually. It seems that `QNetworkAccessManager::post` when taking in a pointer of `QIODevice`, [will try to create another `QSharedPointer`](https://code.woboq.org/qt5/qtbase/src/corelib/io/qnoncontiguousbytedevice.cpp.html#_ZN31QNonContiguousByteDeviceFactory12createSharedEP9QIODevice). This causes problems, because in our case, the pointer to `QIODevice` is already a `QSharedPointer`. Somewhere this causes a segfault because of a double-free. 

I was able to reproduce the issue. The following code is based on the stack trace from gdb.

Here's the code:
```cpp
#include <QBuffer>
#include <QFile>
#include <QPointer>
#include <iostream>
#include <memory>

static void not_deleter(QObject *obj)
{
}

static void deleter(QObject *obj)
{
    delete obj;
}

auto foo(QIODevice *p)
{
    std::cerr << "Before cast\n";
    QBuffer *buf = qobject_cast<QBuffer*>(p);
    std::cerr << "After cast\n";

    std::cerr << "Before unique_ptr ctor\n";
    QSharedPointer<QIODevice>{p, &deleter};      // segfault
    QSharedPointer<QIODevice>{p, &not_deleter};  // OK
    std::cerr << "After unique_ptr ctor\n";
}

int main()
{
    std::cerr << "Begin main()\n";
    QSharedPointer<QFile> ptr{new QFile{}};

    std::cerr << "Before call to foo()\n";
    foo(ptr.data());
    std::cerr << "After call to foo()\n";

    std::cerr << "End main()\n";
    return 0;
}
```

Here's the stack trace:
```
Thread 1 "nheko" received signal SIGSEGV, Segmentation fault.
0x0000000e00000001 in ?? ()
(gdb) bt
#0  0x0000000e00000001 in  ()
#1  0x00007ffff5ea1b29 in QMetaObject::cast(QObject const*) const ()
    at /usr/lib/libQt5Core.so.5
#2  0x00007ffff5db9e7d in QNonContiguousByteDeviceFactory::createShared(QIODevice*) ()
    at /usr/lib/libQt5Core.so.5
#3  0x00007ffff6b39ecc in  () at /usr/lib/libQt5Network.so.5
#4  0x00007ffff6b40330 in  () at /usr/lib/libQt5Network.so.5
#5  0x00007ffff6b41c0c in  () at /usr/lib/libQt5Network.so.5
#6  0x00007ffff6b41cf5 in  () at /usr/lib/libQt5Network.so.5
#7  0x00007ffff6bf4fda in  () at /usr/lib/libQt5Network.so.5
#8  0x00007ffff5ec9452 in QObject::event(QEvent*) () at /usr/lib/libQt5Core.so.5
#9  0x00007ffff76d1e3c in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()
    at /usr/lib/libQt5Widgets.so.5
#10 0x00007ffff76d9816 in QApplication::notify(QObject*, QEvent*) ()
    at /usr/lib/libQt5Widgets.so.5
#11 0x00007ffff5e981e0 in QCoreApplication::notifyInternal2(QObject*, QEvent*) ()
    at /usr/lib/libQt5Core.so.5
#12 0x00007ffff5e9ae46 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () at /usr/lib/libQt5Core.so.5
#13 0x00007ffff5ef50a4 in  () at /usr/lib/libQt5Core.so.5
#14 0x00007ffff2670270 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#15 0x00007ffff2671f69 in  () at /usr/lib/libglib-2.0.so.0
#16 0x00007ffff2671fae in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#17 0x00007ffff5ef4691 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt5Core.so.5
#18 0x00007fffe8f2c152 in  () at /usr/lib/libQt5XcbQpa.so.5
#19 0x00007ffff5e9682b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) ()
    at /usr/lib/libQt5Core.so.5
#20 0x00007ffff5e9fb18 in QCoreApplication::exec() () at /usr/lib/libQt5Core.so.5
#21 0x0000555555a21ee7 in main(int, char**) (argc=1, argv=0x7fffffffe028)
    at /home/chris/dev/nheko/src/main.cc:166
```
See the link above for the source code of `QNonContiguousByteDeviceFactory::createShared`.

Ralith: Referring to [the documentation of `QSharedPointer::create`](https://doc.qt.io/qt-5/qsharedpointer.html#create), you can see that it does not create a new `QSharePointer` owning exactly the same object, but rather creates a new, unrelated `QSharedPointer` with its own object. This makes sense, because (although the weird syntax highlighting makes it a bit hard to see) the value being passed in is a pointer to the pointer, not the pointer itself.

[The documentation for `QNetworkAccessManager::post`](https://doc.qt.io/qt-5/qnetworkaccessmanager.html#post) specifies that you must ensure the `QIODevice` remains valid until the returned reply emits its `finished` signal. If you're seeing memory errors, this may be the cause. I believe a convenient way to ensure the `QIODevice` lives that long would be attaching a closure that captures the `QSharedPointer` that owns it to the signal.

mujx:
Could you remove the full path from the default name on the modal and only set the name of the file?

src/MatrixClient.cc
christarazi: Ah, thanks. Yes that fixed the issue.

src/TextInputWidget.cc
christarazi: Oops I forgot to address this one. I will make the change locally, and add it to the next set of changes if there is one.

christarazi:
New commit added addressing the second round of issues. Thanks again for your vigilance and patience!

mujx:
Thanks again for your work! Could you rebase on master to resolve the conflicts and let the ci run?

christarazi:
I have rebased the changes onto master.

src/timeline/widgets/ImageItem.cc
```diff
@@ -83,7 +84,16 @@ ImageItem::ImageItem(QSharedPointer<MatrixClient> client,
...
 
-        setImage(QPixmap(filename));
+        auto null  = data.isNull();
+        auto reset = data->reset();
+        QPixmap p;
+
+        if (!null && reset) {
+                p.loadFromData(data->readAll());
+        } else {
+                p.load(filename);
```
Ralith: Because `filename` no longer identifies a path on disk, this fallback doesn't make sense. An error message (see [`QIODevice::errorString`](https://doc.qt.io/qt-5/qiodevice.html#errorString) should be displayed instead.

src/timeline/widgets/AudioItem.cc
```diff
@@ -89,14 +89,16 @@ AudioItem::AudioItem(QSharedPointer<MatrixClient> client,
 
 AudioItem::AudioItem(QSharedPointer<MatrixClient> client,
                      const QString &url,
+                     const QSharedPointer<QIODevice> data,
                      const QString &filename,
                      QWidget *parent)
   : QWidget(parent)
   , url_{url}
-  , text_{QFileInfo(filename).fileName()}
+  , text_{QFileInfo{filename}.fileName()}
```
Ralith: `filename` no longer identifies a path on disk.

src/TextInputWidget.cc
```diff
@@ -101,6 +111,43 @@ FilteredTextEdit::keyPressEvent(QKeyEvent *event)
         }
 }
 
+bool
+FilteredTextEdit::canInsertFromMimeData(const QMimeData *source) const
+{
+        return (source->hasImage() || source->hasText() ||
```
Ralith: Isn't `hasText` redundant to checks made by `QTextEdit::canInsertFromMimeData`?

src/MatrixClient.cc
```diff
@@ -1021,20 +1027,19 @@ MatrixClient::makeUploadRequest(const QString &filename)
...
                 return nullptr;
         }
 
         QMimeDatabase db;
-        QMimeType mime = db.mimeTypeForFile(filename, QMimeDatabase::MatchContent);
+        QMimeType mime = db.mimeTypeForData(iodev.data());
 
         QNetworkRequest request(QString(endpoint.toEncoded()));
-        request.setHeader(QNetworkRequest::ContentLengthHeader, file.size());
+        request.setHeader(QNetworkRequest::ContentLengthHeader, iodev->size());
```
Ralith: Per [the `QIODevice::size` documentation](https://doc.qt.io/qt-5/qiodevice.html#size), this is incorrect if the device is not a random access device. The header should be omitted entirely if [`QIODevice::isSequential`](https://doc.qt.io/qt-5/qiodevice.html#isSequential) returns true.

[The spec](https://matrix.org/docs/spec/r0.0.0/client_server.html#post-matrix-media-r0-upload) doesn't seem to require the header at all, so I'm not sure why we're providing it in the first place. I believe it works fine without, so maybe it would be better to just remove this line entirely? Or is it of use to middleboxes sometimes?

src/MatrixClient.cc
```diff
@@ -1021,20 +1027,19 @@ MatrixClient::makeUploadRequest(const QString &filename)
         endpoint.setPath(mediaApiUrl_ + "/upload");
         endpoint.setQuery(query);
 
-        QFile file(filename);
-        if (!file.open(QIODevice::ReadWrite)) {
-                qDebug() << "Error while reading" << filename;
+        if (!iodev->open(QIODevice::ReadOnly)) {
+                qDebug() << "Error while reading buffer" << iodev->errorString();
```
Ralith: A `QIODevice` is not necessarily a buffer.

src/timeline/widgets/VideoItem.cc
```diff
@@ -66,13 +66,15 @@ VideoItem::VideoItem(QSharedPointer<MatrixClient> client,
 
 VideoItem::VideoItem(QSharedPointer<MatrixClient> client,
                      const QString &url,
+                     const QSharedPointer<QIODevice> data,
                      const QString &filename,
                      QWidget *parent)
   : QWidget(parent)
   , url_{url}
   , text_{QFileInfo(filename).fileName()}
```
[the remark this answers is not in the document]
christarazi: For `VideoItem`, `AudioItem`, and `FileItem`, `filename` does actually identify a path on disk. Only `ImageItem`'s doesn't. This is because they are only created by the user explicitly selecting them to upload. 

Once they can be uploaded through other means (copy-paste), then I think it would make sense to modify this (similar to how `ImageItem` is now).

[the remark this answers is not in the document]
Ralith: If they need a uniform interface, then the interface should be semantically uniform. That means no inconsistent special meanings. `filename` should be standardized as purely UI.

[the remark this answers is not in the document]
mujx: I'm also not a fan of this but it could be handled in a different PR.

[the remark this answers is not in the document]
christarazi: I also don't like it, but I thought it would be out of the scope of this PR to add changes unrelated to image pasting. I can definitely tackle this after this PR is merged.

include/timeline/TimelineView.h
```diff
@@ -59,7 +59,8 @@ struct PendingMessage
           , filename(filename)
           , event_id(event_id)
           , widget(widget)
-        {}
+        {
```
mujx: I think those were fine. There are some issues with the clang-format running on travis (macos) and the one on liinux, so this will fail on travis. Not sure if this behaviour can be changed with a flag.

src/timeline/widgets/VideoItem.cc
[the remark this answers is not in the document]
Ralith: This PR changes the pervasive `filename` argument to file-ish widget constructors from "always a full path on disk" to not. IMO is is therefore in scope; inconsistency here is a landmine.

include/timeline/TimelineView.h
christarazi: OK I will revert these lines.

src/timeline/widgets/VideoItem.cc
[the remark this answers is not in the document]
mujx: I merged it before addressing the `filename` issue because I didn't want to drag this PR any further with refactoring. Although it is related, I think it would be better to have a more focused PR without the fear of moving too much out of scope.

christarazi:
I will begin work immediately on addressing the `filename` issue as discussed. Thanks for merging!

alphapapa:
@christarazi Thanks for your work on this!  I just pulled and built and tested, and it worked great!  Using Ubuntu 14.04 with KDE 4, I took a screenshot with KSnapshot, copied it to the clipboard, and pasted it directly into nheko.

@mujx @Ralith Thanks for your work on this as well!

PR #195, opened 2018-01-06
Communities

vurpo:
I made a new pull request for this one because I had branch confusion, and now I merged in the new changes on the original branch instead of communities-merged. (It doesn't seem like you can change the source branch for a PR on GitHub.)


mujx:
Could you take a look at the ci failures? Also the `.vscode` dir should be excluded.

src/MatrixClient.cc
```diff
@@ -315,20 +314,20 @@ MatrixClient::sendRoomMessage(mtx::events::MessageType ty,
 
                 if (!json.isObject()) {
                         qDebug() << "Send message response is not a JSON object";
-                        emit messageSendFailed(roomid, txnId);
```
mujx: We need to keep this.

src/MatrixClient.cc
```diff
@@ -315,20 +314,20 @@ MatrixClient::sendRoomMessage(mtx::events::MessageType ty,
...
                         return;
                 }
 
                 auto object = json.object();
 
                 if (!object.contains("event_id")) {
                         qDebug() << "SendTextMessage: missing event_id from response";
-                        emit messageSendFailed(roomid, txnId);
```
mujx: We also need this.

src/MatrixClient.cc
```diff
@@ -315,20 +314,20 @@ MatrixClient::sendRoomMessage(mtx::events::MessageType ty,
...
                         return;
                 }
 
                 emit messageSent(object.value("event_id").toString(), roomid, txnId);
         });
+
+        incrementTransactionId();
```
mujx: This can be removed.

src/MainWindow.cc
```diff
@@ -128,6 +128,15 @@ MainWindow::MainWindow(QWidget *parent)
         }
 }
 
+void
```
mujx: This is old code and can be removed.

include/MainWindow.h
```diff
@@ -47,6 +47,7 @@ class MainWindow : public QMainWindow
 
 protected:
         void closeEvent(QCloseEvent *event);
+        void keyPressEvent(QKeyEvent *event);
```
mujx: This is old code and can be removed.

src/MatrixClient.cc
vurpo: Yeah, git was very unhelpful when it showed the merge conflict for this file

PR #196, opened 2018-01-09
Add make install and make uninstall rules to cmake

PureTryOut:


mujx:
Could you move the `cmake_uninstall` file in the `cmake` directory?

CMakeLists.txt
```diff
@@ -337,3 +339,24 @@ else()
     add_executable (nheko ${OS_BUNDLE} ${NHEKO_DEPS})
     target_link_libraries (nheko ${NHEKO_LIBS} Qt5::Multimedia)
 endif()
+
+# Install target
```
mujx: Could you wrap this whole section in a `if (LINUX) ... endif(LINUX)` block to fix the build on macos? Any other fix is also welcome.

PureTryOut:
I used a `if(UNIX AND NOT APPLE)` block instead of `if(LINUX)` as with the latter, the install rule was not added to the makefile.

mujx:
Dank je!

PR #198, opened 2018-01-10
filter content in sync that is currently unhandled

krombel:
I had a look at sync.cpp and checked which parts of the sync response
are currently handled and which not.
As I think it is unneccessary to let the unhandled data be transmitted
without beeing handled I added these filters.

In the same term I increased the timeout server-side to 30s as Riot
defaults to this value as well.
Especially now when a lots of presence-updates are not send anymore
this value is more relevant

### DIFF
--- a/include/MatrixClient.h
+++ b/include/MatrixClient.h
@@ -159,4 +159,7 @@ private:
 
         // Token to be used for the next sync.
         QString next_batch_;
+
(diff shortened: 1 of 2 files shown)


mujx:
Could you apply the same filter to `MatrixClient::initialSync` too?

PR #220, opened 2018-01-27
 Support audio, video, generic file for pasting 

christarazi:
This PR allows for pasting and uploading all file types, including support for audio and video.

### DIFF
--- a/include/TextInputWidget.h
+++ b/include/TextInputWidget.h
@@ -53,6 +53,7 @@ signals:
         void heightChanged(int height);
         void startedTyping();
         void stoppedTyping();
+        void startedUpload();
         void message(QString);
         void command(QString name, QString args);
         void image(QSharedPointer<QIODevice> data, const QString &filename);
--- a/include/TextInputWidget.h
+++ b/include/TextInputWidget.h
@@ -27,12 +27,12 @@
 #include "FlatButton.h"
 #include "LoadingIndicator.h"
 
-#include "dialogs/PreviewImageOverlay.h"
+#include "dialogs/PreviewUploadOverlay.h"
 
 #include "emoji/PickButton.h"
 
 namespace dialogs {
-class PreviewImageOverlay;
+class PreviewUploadOverlay;
 }
 
 class FilteredTextEdit : public QTextEdit
@@ -70,7 +70,7 @@ private:
         size_t history_index_;
         QTimer *typingTimer_;
 
-        dialogs::PreviewImageOverlay previewDialog_;
+        dialogs::PreviewUploadOverlay previewDialog_;
 
         void textChanged();
         void uploadData(const QByteArray data, const QString &media, const QString &filename);
--- a/include/TextInputWidget.h
+++ b/include/TextInputWidget.h
@@ -75,10 +75,7 @@ private:
         void textChanged();
         void uploadData(const QByteArray data, const QString &media, const QString &filename);
         void afterCompletion(int);
-        void showPreview(const QMimeData *source,
-                         const QStringList &formats,
-                         const QString &media,
-                         const QString &type);
+        void showPreview(const QMimeData *source, const QStringList &formats);
 };
 
 class TextInputWidget : public QWidget
--- a/include/TextInputWidget.h
+++ b/include/TextInputWidget.h
@@ -55,7 +55,10 @@ signals:
         void stoppedTyping();
         void message(QString);
         void command(QString name, QString args);
-        void image(const QSharedPointer<QIODevice> iodev, const QString &img_name);
+        void image(const QSharedPointer<QIODevice> data, const QString &filename);
+        void audio(const QSharedPointer<QIODevice> data, const QString &filename);
+        void video(const QSharedPointer<QIODevice> data, const QString &filename);
+        void file(const QSharedPointer<QIODevice> data, const QString &filename);
 
 protected:
         void keyPressEvent(QKeyEvent *event) override;
@@ -70,8 +73,12 @@ private:
         dialogs::PreviewImageOverlay previewDialog_;
 
         void textChanged();
-        void receiveImage(const QByteArray img, const QString &img_name);
+        void uploadData(const QByteArray data, const QString &media, const QString &filename);
         void afterCompletion(int);
+        void showPreview(const QMimeData *source,
+                         const QStringList &formats,
+                         const QString &media,
+                         const QString &type);
 };
 
 class TextInputWidget : public QWidget
@@ -95,9 +102,10 @@ signals:
         void sendTextMessage(QString msg);
         void sendEmoteMessage(QString msg);
 
-        void uploadImage(QSharedPointer<QIODevice> data, const QString &filename);
-        void uploadFile(QSharedPointer<QIODevice> data, const QString &filename);
-        void uploadAudio(QSharedPointer<QIODevice> data, const QString &filename);
+        void uploadImage(const QSharedPointer<QIODevice> data, const QString &filename);
+        void uploadFile(const QSharedPointer<QIODevice> data, const QString &filename);
+        void uploadAudio(const QSharedPointer<QIODevice> data, const QString &filename);
+        void uploadVideo(const QSharedPointer<QIODevice> data, const QString &filename);
 
         void sendJoinRoomRequest(const QString &room);
 
--- a/include/TextInputWidget.h
+++ b/include/TextInputWidget.h
@@ -55,10 +55,10 @@ signals:
         void stoppedTyping();
         void message(QString);
         void command(QString name, QString args);
-        void image(const QSharedPointer<QIODevice> data, const QString &filename);
-        void audio(const QSharedPointer<QIODevice> data, const QString &filename);
-        void video(const QSharedPointer<QIODevice> data, const QString &filename);
-        void file(const QSharedPointer<QIODevice> data, const QString &filename);
+        void image(QSharedPointer<QIODevice> data, const QString &filename);
+        void audio(QSharedPointer<QIODevice> data, const QString &filename);
+        void video(QSharedPointer<QIODevice> data, const QString &filename);
+        void file(QSharedPointer<QIODevice> data, const QString &filename);
 
 protected:
         void keyPressEvent(QKeyEvent *event) override;
--- a/src/MatrixClient.cc
+++ b/src/MatrixClient.cc
@@ -280,7 +280,8 @@ MatrixClient::sendRoomMessage(mtx::events::MessageType ty,
                               int txnId,
                               const QString &roomid,
                               const QString &msg,
-                              const QSharedPointer<QIODevice> media,
+                              const QString &mime,
+                              const qint64 media_size,
                               const QString &url) noexcept
 {
         QUrlQuery query;
@@ -291,28 +292,8 @@ MatrixClient::sendRoomMessage(mtx::events::MessageType ty,
                          QString("/rooms/%1/send/m.room.message/%2").arg(roomid).arg(txnId));
         endpoint.setQuery(query);
 
-        QString msgType("");
-
-        QMimeDatabase db;
-        QMimeType mime;
-        qint64 m_size = 0;
-
-        if (!media.isNull()) {
-                // Must seek to beginning of device because mimeTypeForData() does not
-                // automatically.
-                if (!media->reset()) {
-                        qWarning()
-                          << "Failed to seek to beginning of device:" << media->errorString()
-                          << '\n'
-                          << "Sending message with inaccurate MIME type...";
-                }
-
-                mime   = db.mimeTypeForData(media.data());
-                m_size = media->size();
-        }
-
         QJsonObject body;
-        QJsonObject info = {{"size", m_size}, {"mimetype", mime.name()}};
+        QJsonObject info = {{"size", media_size}, {"mimetype", mime}};
 
         switch (ty) {
         case mtx::events::MessageType::Text:
@@ -829,68 +810,97 @@ MatrixClient::messages(const QString &roomid, const QString &from_token, int lim
 
 void
 MatrixClient::uploadImage(const QString &roomid,
-                          const QSharedPointer<QIODevice> data,
-                          const QString &filename)
+                          const QString &filename,
+                          const QSharedPointer<QIODevice> data)
 {
         auto reply = makeUploadRequest(data);
 
         if (reply == nullptr)
                 return;
 
-        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, data, filename]() {
-                auto object = upload(reply);
-                if (object.isEmpty())
+        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, filename, data]() {
+                auto json = getUploadReply(reply);
+                if (json.isEmpty())
                         return;
 
-                emit imageUploaded(roomid, data, filename, object.value("content_uri").toString());
+                auto mime = reply->request().header(QNetworkRequest::ContentTypeHeader).toString();
+                auto size =
+                  reply->request().header(QNetworkRequest::ContentLengthHeader).toLongLong();
+
+                emit imageUploaded(
+                  roomid, filename, json.value("content_uri").toString(), mime, size);
         });
 }
 
 void
 MatrixClient::uploadFile(const QString &roomid,
-                         const QSharedPointer<QIODevice> data,
-                         const QString &filename)
+                         const QString &filename,
+                         const QSharedPointer<QIODevice> data)
 {
         auto reply = makeUploadRequest(data);
 
-        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, data, filename]() {
-                auto object = upload(reply);
-                if (object.isEmpty())
+        if (reply == nullptr)
+                return;
+
+        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, filename, data]() {
+                auto json = getUploadReply(reply);
+                if (json.isEmpty())
                         return;
 
-                emit fileUploaded(roomid, data, filename, object.value("content_uri").toString());
+                auto mime = reply->request().header(QNetworkRequest::ContentTypeHeader).toString();
+                auto size =
+                  reply->request().header(QNetworkRequest::ContentLengthHeader).toLongLong();
+
+                emit fileUploaded(
+                  roomid, filename, json.value("content_uri").toString(), mime, size);
         });
 }
 
 void
 MatrixClient::uploadAudio(const QString &roomid,
-                          const QSharedPointer<QIODevice> data,
-                          const QString &filename)
+                          const QString &filename,
+                          const QSharedPointer<QIODevice> data)
 {
         auto reply = makeUploadRequest(data);
 
-        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, data, filename]() {
-                auto object = upload(reply);
-                if (object.isEmpty())
+        if (reply == nullptr)
+                return;
+
+        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, filename, data]() {
+                auto json = getUploadReply(reply);
+                if (json.isEmpty())
                         return;
 
-                emit audioUploaded(roomid, data, filename, object.value("content_uri").toString());
+                auto mime = reply->request().header(QNetworkRequest::ContentTypeHeader).toString();
+                auto size =
+                  reply->request().header(QNetworkRequest::ContentLengthHeader).toLongLong();
+
+                emit audioUploaded(
+                  roomid, filename, json.value("content_uri").toString(), mime, size);
         });
 }
 
 void
 MatrixClient::uploadVideo(const QString &roomid,
-                          const QSharedPointer<QIODevice> data,
-                          const QString &filename)
+                          const QString &filename,
+                          const QSharedPointer<QIODevice> data)
 {
         auto reply = makeUploadRequest(data);
 
-        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, data, filename]() {
-                auto object = upload(reply);
-                if (object.isEmpty())
+        if (reply == nullptr)
+                return;
+
+        connect(reply, &QNetworkReply::finished, this, [this, reply, roomid, filename, data]() {
+                auto json = getUploadReply(reply);
+                if (json.isEmpty())
                         return;
 
-                emit videoUploaded(roomid, data, filename, object.value("content_uri").toString());
+                auto mime = reply->request().header(QNetworkRequest::ContentTypeHeader).toString();
+                auto size =
+                  reply->request().header(QNetworkRequest::ContentLengthHeader).toLongLong();
+
+                emit videoUploaded(
+                  roomid, filename, json.value("content_uri").toString(), mime, size);
         });
 }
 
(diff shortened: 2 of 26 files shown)


mujx:
Travis is failing for Qt 5.7. Making the `nullptr` argument optional might help.

src/TextInputWidget.cc
```diff
@@ -135,28 +135,23 @@ FilteredTextEdit::canInsertFromMimeData(const QMimeData *source) const
...
+                // Generic file for any platform.
                 // Special case for X11 users. See "Notes for X11 Users" in source.
                 // Source: http://doc.qt.io/qt-5/qclipboard.html
-                previewDialog_.setImageAndCreate(source->text());
-                previewDialog_.show();
+                previewDialog_.setPreview(source->text());
+        } else if (!audio.empty()) {
+                showPreview(source, audio, "audio", "flac");
+        } else if (!video.empty()) {
+                showPreview(source, video, "video", "webm");
```
Ralith: Why bother to pass in a default media type, when the passed-in formats list in every call is both non-empty and contain only appropriate values? It should be safe to e.g. always select the first.

src/TextInputWidget.cc
```diff
@@ -233,11 +228,40 @@ FilteredTextEdit::textChanged()
...
+}
+
+void
+FilteredTextEdit::showPreview(const QMimeData *source,
+                              const QStringList &formats,
+                              const QString &media,
+                              const QString &type)
+{
+        const auto idx = formats.indexOf(
+          QRegularExpression{media + "/.+", QRegularExpression::CaseInsensitiveOption});
```
Ralith: Do we really need to search again, when in practice `formats` always contains nothing but matches and is nonempty?

src/TextInputWidget.cc
```diff
@@ -233,11 +228,40 @@ FilteredTextEdit::textChanged()
...
+
+        // Note: in the future we may want to look into what the best choice is from the
+        // formats list. For now we will default to |type| format.
+        QString mime = media + '/' + type;
```
Ralith: This should never be needed, given the current uses of this function.

src/Utils.cc
```diff
@@ -133,3 +133,19 @@ utils::firstChar(const QString &input)
 
         return QString::fromUcs4(&input.toUcs4().at(0), 1).toUpper();
 }
+
+QString
+utils::humanReadableFileSize(const qlonglong bytes)
+{
+        constexpr static const char *units[] = {"B", "KB", "MB", "GB", "TB"};
```
Ralith: Use the correct [IEEE 1541](https://en.wikipedia.org/wiki/IEEE_1541-2002) units: KiB, MiB, GiB, TiB, etc

src/Utils.cc
```diff
@@ -133,3 +133,19 @@ utils::firstChar(const QString &input)
 
         return QString::fromUcs4(&input.toUcs4().at(0), 1).toUpper();
 }
+
+QString
+utils::humanReadableFileSize(const qlonglong bytes)
```
Ralith: `uint64_t`

src/Utils.cc
```diff
@@ -133,3 +133,19 @@ utils::firstChar(const QString &input)
...
+        constexpr static const int length    = sizeof(units) / sizeof(units[0]);
+
+        int u       = 0;
+        double size = static_cast<double>(bytes);
+        while (size >= 1024.0 && u < length) {
+                ++u;
+                size /= 1024.0;
+        }
+
+        return QString::number(size, 'f', 2) + ' ' + units[u];
```
Ralith: Use `'g'` formatting and 4 significant digits. This provides a consistent level of precision and reduces variation in string length, rather than just unconditionally slapping on two decimal places no matter the order of magnitude.

src/dialogs/PreviewUploadOverlay.cc
```diff
@@ -0,0 +1,181 @@
...
+        vlayout->addWidget(&titleLabel_);
+        vlayout->addWidget(&infoLabel_);
+        vlayout->addWidget(&fileName_);
+        vlayout->addLayout(hlayout);
+
+        connect(&upload_, &QPushButton::clicked, [&]() {
+                emit confirmUpload(data_, mediaType_, fileName_.text());
+                close();
+        });
+        connect(&cancel_, &QPushButton::clicked, [&]() { close(); });
```
Ralith: This could connect to `close` directly.

src/dialogs/PreviewUploadOverlay.cc
```diff
@@ -0,0 +1,181 @@
...
+
+void
+PreviewUploadOverlay::init()
+{
+        auto window   = QApplication::activeWindow();
+        auto winsize  = window->frameGeometry().size();
+        auto center   = window->frameGeometry().center();
+        auto img_size = image_.size();
+
+        fileName_.setText(QFileInfo{filePath_}.fileName());
```
Ralith: `filePath_` is not guaranteed to refer to a real file. Do we have a guarantee that `QFileInfo` will remain well-behaved in that case?

src/timeline/widgets/AudioItem.cc
```diff
@@ -94,30 +95,14 @@ AudioItem::AudioItem(QSharedPointer<MatrixClient> client,
                      QWidget *parent)
   : QWidget(parent)
   , url_{url}
-  , text_{QFileInfo{filename}.fileName()}
+  , text_{filename}
   , client_{client}
 {
-        Q_UNUSED(data);
-        readableFileSize_ = calculateFileSize(QFileInfo{filename}.size());
+        readableFileSize_ = utils::humanReadableFileSize(data->size());
```
Ralith: `QIODevice::size` is not guaranteed to be the total size of the data.

src/timeline/widgets/FileItem.cc
```diff
@@ -81,30 +82,14 @@ FileItem::FileItem(QSharedPointer<MatrixClient> client,
                    QWidget *parent)
   : QWidget(parent)
   , url_{url}
-  , text_{QFileInfo{filename}.fileName()}
+  , text_{filename}
   , client_{client}
 {
-        Q_UNUSED(data);
-        readableFileSize_ = calculateFileSize(QFileInfo{filename}.size());
+        readableFileSize_ = utils::humanReadableFileSize(data->size());
```
Ralith: `QIODevice::size` is not guaranteed to be the total size of the data.

src/timeline/widgets/VideoItem.cc
```diff
@@ -71,26 +72,10 @@ VideoItem::VideoItem(QSharedPointer<MatrixClient> client,
                      QWidget *parent)
   : QWidget(parent)
   , url_{url}
-  , text_{QFileInfo(filename).fileName()}
+  , text_{filename}
   , client_{client}
 {
-        Q_UNUSED(data);
-        readableFileSize_ = calculateFileSize(QFileInfo(filename).size());
+        readableFileSize_ = utils::humanReadableFileSize(data->size());
```
Ralith: `QIODevice::size` is not guaranteed to be the total size of the data.

src/TextInputWidget.cc
christarazi: In the previous PR, I recall we were discussing which default image format we should fall-back to. I applied the same concept here. If you feel that having a default fall-back is unnecessary and it is always safe to select the first of the specific formats list (the filtered ones), then I will remove the extra parameters.

src/TextInputWidget.cc
christarazi: I presume this comment relies on the previous comment, but I'll reply anyways.

This function's purpose is to be a helper to search the formats list, and if that fails, have the correct fall-back, and retrieve the data (as QByteArray) and send it on over to PreviewUploadOverlay. It is purely for not having to rewrite this logic in every if-statement.

src/Utils.cc
christarazi: Noted and corrected. Thanks!

src/dialogs/PreviewUploadOverlay.cc
christarazi: Good catch!

src/dialogs/PreviewUploadOverlay.cc
christarazi: Yes it is guaranteed. [`QFileInfo` does not require the file to exist](http://doc.qt.io/qt-5.10/qfileinfo.html#exists).

src/TextInputWidget.cc
Ralith: This code as written will *never* use the "defaults" for anything. It's not clear to me what problem they're intended to solve. Dead code is a maintenance hazard.

src/TextInputWidget.cc
Ralith: The formats list that you pass in seems to always have already been filtered such that this search is meaningless, and a fallback will never occur.

src/TextInputWidget.cc
christarazi: OK I will do without it. Thanks for your attention to detail!

src/timeline/widgets/AudioItem.cc
christarazi: Fixed in commit f159b7b.

src/timeline/widgets/FileItem.cc
christarazi: Fixed in commit f159b7b.

src/timeline/widgets/VideoItem.cc
christarazi: Fixed in commit f159b7b.

src/MatrixClient.cc
```diff
@@ -294,11 +294,25 @@ MatrixClient::sendRoomMessage(mtx::events::MessageType ty,
...
+                // automatically.
+                if (!media->reset()) {
+                        qWarning()
+                          << "Failed to seek to beginning of device:" << media->errorString()
+                          << '\n'
+                          << "Sending message with inaccurate MIME type...";
+                }
+
+                mime   = db.mimeTypeForData(media.data());
+                m_size = media->size();
```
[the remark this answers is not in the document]
christarazi: Fixed in commit f159b7b.

christarazi:
I made a big refactoring change in commit f159b7b because @Ralith pointed out something very important -- `QIODevice` does not need to be passed around everywhere. The information from it can be retrieved once and those can be passed instead. 

This eliminates any future potential issues with trying to upload data from different `QIODevice`s that are sequential. 

@mujx Also, I hope you don't mind me fixing issue #228 inside this PR. The changes in this PR are very much interlinked with that issue, so I thought it would be cleaner to just include it in here.

Thanks again for your time in reviewing this!

mujx:
I had a crash while pasting a png image.

```
    frame #6: 0x000000010025cf1e nheko`QList<QString>::first(this=0x00007ffeefbfd6d0) const at qlist.h:347
    frame #7: 0x000000010025c335 nheko`FilteredTextEdit::showPreview(this=0x000000010fc75dd0, source=0x0000000126c2c2e0, formats=0x00007ffeefbfd6d0) at TextInputWidget.cc:250
    frame #8: 0x000000010025bef2 nheko`FilteredTextEdit::insertFromMimeData(this=0x000000010fc75dd0, source=0x0000000126c2c2e0) at TextInputWidget.cc:144
    frame #9: 0x00000001012c354e QtWidgets`___lldb_unnamed_symbol2858$$QtWidgets + 1134
    frame #10: 0x00000001012c2190 QtWidgets`QWidgetTextControl::processEvent(QEvent*, QPointF const&, QWidget*) + 80
    frame #11: 0x00000001012aeffc QtWidgets`QTextEdit::keyPressEvent(QKeyEvent*) + 604
    frame #12: 0x000000010025b6f0 nheko`FilteredTextEdit::keyPressEvent(this=0x000000010fc75dd0, event=0x00007ffeefbfdf58) at TextInputWidget.cc:124
```
There should be check there if the list is empty.

christarazi:
@mujx Did you do anything out of the ordinary to cause this? 

According to @Ralith's previous comments, he mentions that Qt guarantees a valid MIME type. Hence why the list in not checked. If the list were to be checked, then this begs the question of: do we need some sort of fall back (which I previously implemented, but removed due to the code review comments)?

Ralith:
In the "Detailed Description" section of [the QMimeData documentation](https://doc.qt.io/qt-5/qmimedata.html#QMimeData), `hasImage` is documented as corresponding to a mime type of `image/*`. Despite this, the implementation appears to actually treat it as `application/x-qt-image`, presumably to indicate that there's actually a `QVariant` containing a `QImage` attached to it rather than encoded image data as is present under normal circumstances. This probably accounts for the failure.

@christarazi did you actually test the current implementation and find it working? If so, I wonder why sometimes you get real image data and sometimes you get this weird magic Qt thing.

The removed code was not a functional fallback, because requesting e.g. `image/png` data from a `QMimeData` that does not contain any will not accomplish anything but crashing or unexpected behavior in a different location.

christarazi:
@Ralith Yes I did test the current implementation. I used it as my daily driver and would rebase with master practically daily, each time ensuring that pasting audio, video, images and generic files work.

I have yet to come across something that would cause a crash. Which is why I'm particularly interested in what @mujx did to cause this.

Ralith:
At a guess, perhaps copying images from within the same (Qt) program produces that type?

christarazi:
Do you mean copying an image from nheko itself? I don't think that is supported.

mujx:
@christarazi The crash mentioned happened on macOS. In the first PR about the 'paste image` feature there was a problem were macOS (or qt on mac) would not paste the actual image but a generic image placeholder. Probably that's what causing this.

In any case it would better for the app to do nothing than crash, so I think the check is necessary there.

christarazi:
@mujx Ah I see. I can add a check and log a message if the list is empty then. 

I'm curious to find out what MIME type macOS is using though. With the MIME type, maybe we can discover a better way to handle this. 

On February 15, 2018 12:11:02 AM PST, mujx <notifications@github.com> wrote:
>@christarazi The crash mentioned happened on macOS. In the first PR
>about the 'paste image` feature there was a problem were macOS (or qt
>on mac) would not paste the actual image but a generic image
>placeholder. Probably that's what causing this.
>
>In any case it would better for the app to do nothing than crash, so I
>think the check is necessary there.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/mujx/nheko/pull/220#issuecomment-365853115

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Ralith:
We know `hasImage` is succeeding, so `application/x-qt-image` must be at least one of the types. Perhaps there's another type that actually contains useful data, however; the full list would be interesting to see.

mujx:
`QMimeData::formats` returns the following for a pasted png image.
```
("text/uri-list", "text/plain", "application/x-qt-image")
```

christarazi:
Interesting...even though it's an image, Qt still prioritizes `text/plain` and does not have an entry that starts with `image/`, which would explain the crash. Are you copying an image file from the macOS file explorer? That would probably mean that the `text/plain` is the path to the file.

What's annoying about this is that, as @Ralith mentions, the `QMimeData` documentation seems to be inaccurate, or at least too vague.

We probably need to think of a more general way of filtering the list.

Ralith:
Any reason not to just treat images the same way we do video and etc., i.e. check `image.empty()` instead of `hasImage`?

I bet `text/uri-list` has the path to the original in some form, so special-casing that would be nice for futureu work.

mujx:
@Ralith The `text/uri-list` has a `QUrl` with the file path (can be accessed through `QMimeData::urls()`) and `QMimeData::text()` returns the filname. I could work on adding that special case but I'm not sure how the if-else order should look.

Ralith:
I'd say if the `QMimeData` contains a single `file:` URL to a local file in addition to any other datatypes, the URL should take precedence. Care will be required to gracefully handle inputs with multiple URLs, and with non-`file:` URLs.

christarazi:
How does the file URL help in determining what type of file it is? That's the challenge I see. The obvious solution is to map file extensions to a MIME type. Otherwise, adding a special case to just upload the file (without determining its type), will not create the corresponding timeline widget (image preview, etc.).

christarazi:
Ah yes that should work perfectly.

src/MatrixClient.cc
```diff
@@ -1227,3 +1198,39 @@ MatrixClient::makeUploadRequest(QSharedPointer<QIODevice> iodev)
...
+        if (!json.isObject()) {
+                qDebug() << "Media upload: Response is not a json object.";
+                return object;
+        }
+
+        object = json.object();
+        if (!object.contains("content_uri")) {
+                qDebug() << "Media upload: Missing content_uri key";
+                qDebug() << object;
+                return object;
```
Ralith: Returning a nonempty object that isn't actually usable will probably lead to bugs down the line.

src/TextInputWidget.cc
```diff
@@ -135,28 +135,42 @@ FilteredTextEdit::canInsertFromMimeData(const QMimeData *source) const
...
+
+        if (!image.empty()) {
+                showPreview(source, image);
+        } else if (!audio.empty()) {
+                showPreview(source, audio);
+        } else if (!video.empty()) {
+                showPreview(source, video);
+        } else if (source->hasUrls()) {
+                QString path;
+                if (QFileInfo{source->text()}.exists()) {
```
Ralith: Why is this branch needed? Shouldn't processing the URLs directly work in all cases where this would?

src/TextInputWidget.cc
```diff
@@ -135,28 +135,42 @@ FilteredTextEdit::canInsertFromMimeData(const QMimeData *source) const
...
-        } else if (source->hasFormat("x-special/gnome-copied-files") &&
-                   QImageReader{source->text()}.canRead()) {
+                if (!path.isEmpty() && QFileInfo{path}.exists()) {
+                        previewDialog_.setPreview(path);
+                } else {
+                        qWarning()
+                          << "Clipboard does not contain any valid file paths:" << source->urls();
+                }
+        } else if ((source->hasText() || source->hasFormat("x-special/gnome-copied-files")) &&
+                   QFileInfo{source->text()}.exists()) {
```
Ralith: What prevents this from triggering if I paste plain text that happens to be a path to a local file?

christarazi: Ah good catch. That was an oversight.

src/TextInputWidget.cc
christarazi: It's there to short-circuit the rest of the code in case `source` contains URLs and also a plain version of the file path (without the `file://` prefix).

include/TextInputWidget.h
```diff
@@ -55,7 +55,10 @@ class FilteredTextEdit : public QTextEdit
         void stoppedTyping();
         void message(QString);
         void command(QString name, QString args);
-        void image(const QSharedPointer<QIODevice> iodev, const QString &img_name);
+        void image(const QSharedPointer<QIODevice> data, const QString &filename);
+        void audio(const QSharedPointer<QIODevice> data, const QString &filename);
+        void video(const QSharedPointer<QIODevice> data, const QString &filename);
+        void file(const QSharedPointer<QIODevice> data, const QString &filename);
```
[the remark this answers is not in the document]
christarazi: That's a remnant of my coding style, I try to be const-correct as much as possible. I can remove them.

src/MatrixClient.cc
christarazi: Good catch! This was not my code, I just moved it to its own function because it was being repeated in every file type upload. Will return an empty object instead.

src/TextInputWidget.cc
Ralith: Why is that useful? Processing the URLs is not hard, and is far more likely to be reliable; there's no guarantee that the text is a path.

src/TextInputWidget.cc
```diff
@@ -135,28 +135,42 @@ FilteredTextEdit::canInsertFromMimeData(const QMimeData *source) const
...
-                previewDialog_.show();
+                if (!path.isEmpty() && QFileInfo{path}.exists()) {
+                        previewDialog_.setPreview(path);
+                } else {
+                        qWarning()
+                          << "Clipboard does not contain any valid file paths:" << source->urls();
+                }
         } else if (source->hasFormat("x-special/gnome-copied-files") &&
-                   QImageReader{source->text()}.canRead()) {
+                   QFileInfo{source->text()}.exists()) {
```
Ralith: Why pull the path from `text/plain` (which could be anything) when we've just verified that more stable and strongly typed `x-special/gnome-copied-files` data is available?

src/TextInputWidget.cc
christarazi: It's a premature optimization. On second thought, it is not really that useful. Will change.

src/TextInputWidget.cc
christarazi: Will remove.

src/TextInputWidget.cc
```diff
@@ -165,9 +162,7 @@ FilteredTextEdit::insertFromMimeData(const QMimeData *source)
                         qWarning()
                           << "Clipboard does not contain any valid file paths:" << source->urls();
                 }
-        } else if (source->hasFormat("x-special/gnome-copied-files") &&
-                   QFileInfo{source->text()}.exists()) {
-                // Generic file for any platform.
+        } else if (source->hasFormat("x-special/gnome-copied-files")) {
                 // Special case for X11 users. See "Notes for X11 Users" in source.
                 // Source: http://doc.qt.io/qt-5/qclipboard.html
                 previewDialog_.setPreview(source->text());
```
Ralith: Again: why use the `text/plain` data instead of the `x-special/gnome-copied-files` data?

src/dialogs/PreviewUploadOverlay.cc
```diff
@@ -0,0 +1,181 @@
...
+                }
+                isImage_ = true;
+        } else {
+                auto const info = QString{"Media type: %1\n"
+                                          "Media size: %2\n"}
+                                    .arg(mime)
+                                    .arg(utils::humanReadableFileSize(data.size()));
+
+                titleLabel_.setText(QString{tr(DEFAULT)}.arg("file"));
+                infoLabel_.setText(tr(info.toLocal8Bit()));
```
Ralith: Why are you transcoding this out of unicode, and why are you feeding a dynamic string to `tr`?

src/dialogs/PreviewUploadOverlay.cc
```diff
@@ -0,0 +1,181 @@
...
+                return;
+        }
+
+        auto const &split = mime.name().split('/');
+
+        mediaType_ = split[0];
+        filePath_  = file.fileName();
+        isImage_   = false;
+
+        if (mediaType_ == "image") {
```
Ralith: Everything below here should be factored out into a common function called by both overloads.

mujx:
Great work! Everything works fine on mac. I'll test on windows today.

One thing I noticed is that the spinner doesn't appear when you upload through pasting, like it does when uploading manually. So you might think that something went wrong.

Also could you fix the build on mac (linter fails). I also get the same output with you from the linter, don't why on travis it's different.

src/dialogs/PreviewUploadOverlay.cc
christarazi: Ah good catch!

src/dialogs/PreviewUploadOverlay.cc
christarazi: Will do!

mujx:
You can simply adjust the curly braces [here](https://travis-ci.org/mujx/nheko/jobs/343074232#L2208).

christarazi:
Should be good to go. Thanks for your time @mujx and @Ralith!

PR #224, opened 2018-02-01
Added AppData manifest file for Gnome Software.

xvitaly:
Added AppData manifest file for Gnome Software.

### DIFF
--- a/resources/nheko.appdata.xml
+++ b/resources/nheko.appdata.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 Vitaly Zaitsev <vitaly@easycoding.org> -->
+<component type="desktop">
+  <id>nheko.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPLv3+ and Public Domain and MIT and Boost and ASL 2.0 and CC-BY</project_license>
+  <name>nheko</name>
+  <summary>Desktop client for the Matrix protocol</summary>
+  <description>
+    <p>
+      The motivation behind the project is to provide a native desktop app
+      for Matrix that feels more like a mainstream chat app.
+    </p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://dl.dropboxusercontent.com/s/umojysx3gv80qs9/nheko-default-theme.png</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://github.com/mujx/nheko</url>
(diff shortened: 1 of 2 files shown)


resources/nheko.appdata.xml
```diff
@@ -0,0 +1,23 @@
...
+  <id>nheko.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPLv3+ and Public Domain and MIT and Boost and Apache License and CC-BY-4.0</project_license>
+  <name>nheko</name>
+  <summary>Desktop client for the Matrix protocol</summary>
+  <description>
+    <p>
+      The motivation behind the project is to provide a native desktop app
+      for Matrix that feels more like a mainstream chat app (Riot,
+      Telegram etc) and less like an IRC client.
```
MTRNord: Feels/Sounds like matrix/nheko is a IRC protocol which is isn't. Maybe leave the IRC part out as neither nheko nor Matrix directly has something to do with IRC (and yes it has a bridge and yes both are chat programs)

resources/nheko.appdata.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 Vitaly Zaitsev <vitaly@easycoding.org> -->
+<component type="desktop">
+  <id>nheko.desktop</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPLv3+ and Public Domain and MIT and Boost and Apache License and CC-BY-4.0</project_license>
```
MTRNord: Not sure but shouldn't here be just the license of nheko and not of all of its dependencies?

xvitaly: License of main project and all it's dependencies.

resources/nheko.appdata.xml
xvitaly: I got this information from Readme.md.

MTRNord: Hm ok :) Still leaving it here maybe for mujx to read it (didn't compare to the nheko README.)

mujx: I agree with @MTRNord. Could you use the description from https://github.com/mujx/nheko/blob/master/.ci/linux/create-packages.sh#L55

xvitaly: @mujx Description need to be longer than one row.

mujx: Then I guess you can stop at `chat app`.

xvitaly: @mujx Fixed.

PR #227, opened 2018-02-05
Fix #217: create space for checkmark beforehand

### LINKED ISSUE #217: Text wraps after checkmark gets displayed
### System
- Nheko commit/version: 92a578f Convert 'if' to 'else if' to remove a no-op
- Operating System: Arch Linux x86_64
- Qt version: 5.10.0-3
- C++ compiler: GCC 7.2.1+20171224-2
- Desktop Environment: i3wm

### Actual behavior
After a text has been received by the server, a checkmark appears (extremely useful, thanks for implementing!). When it appears, the text next to it wraps differently than before the checkmark appeared. So for me the line wrapping changes after sending the message and waiting a second or so.

### Expected behavior
Space for the checkmark is reserved, so the text wraps the same way when it is displayed or not yet displayed.

### Steps to reproduce
Send a message.

christarazi:
Allocates a " " (space character) to be a placeholder until the message is verifiably received on the other side. This seems to fix the issue because the checkmark was previously being inserted into the layout at run-time, which would move other push the widgets surrounding it.

### DIFF
--- a/src/timeline/TimelineItem.cc
+++ b/src/timeline/TimelineItem.cc
@@ -61,6 +61,10 @@ TimelineItem::init()
 
         mainLayout_->setContentsMargins(conf::timeline::headerLeftMargin, 0, 0, 0);
         mainLayout_->setSpacing(0);
+
+        checkmark_ = new QLabel(" ", this);
+        checkmark_->setStyleSheet(
+          QString("font-size: %1px;").arg(conf::timeline::fonts::timestamp));
 }
 
 /*
@@ -108,6 +112,7 @@ TimelineItem::TimelineItem(mtx::events::MessageType ty,
                 messageLayout_->addWidget(body_, 1);
         }
 
+        messageLayout_->addWidget(checkmark_);
         messageLayout_->addWidget(timestamp_);
         mainLayout_->addLayout(messageLayout_);
 }
@@ -239,6 +244,7 @@ TimelineItem::TimelineItem(const mtx::events::RoomEvent<mtx::events::msg::Notice
                 messageLayout_->addWidget(body_, 1);
         }
 
+        messageLayout_->addWidget(checkmark_);
         messageLayout_->addWidget(timestamp_);
         mainLayout_->addLayout(messageLayout_);
 }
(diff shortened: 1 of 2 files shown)


mujx:
Thanks a lot for the help with this!

I tried it but the issue is still there. I pasted the following text and it was wrapped when the checkmark appeared.

`
Omnis quod sit aliquam quo quia nam adipisci et. Ipsa nisi et quis rerum perspiciatis numquam ducimus. Consequuntur mollitia animi voluptas non consequuntur. Voluptas molestiae quis quo repellat ex ipsa cum hic. Est officia tenetur ex quo tenetur.
`
Also it needs to be adjusted for the messages with a custom widget (i.e images, files, audio & video).

I think adding a space on the left equal to the width of the check-mark would solve the issue.  The fix fails because the space and the check-mark character don't have the same width (4 vs 11 in my system). `QFontMetrics` can help with this.

christarazi:
Ah, I tried my own sample test but it was evidently not sufficient as shown by yours. I have forced the QLabel to a fixed width, so that should cover all the cases. Also added the missing `addWidget` for images, audio, etc.

src/timeline/TimelineItem.cc
```diff
@@ -61,6 +63,14 @@ TimelineItem::init()
 
         mainLayout_->setContentsMargins(conf::timeline::headerLeftMargin, 0, 0, 0);
         mainLayout_->setSpacing(0);
+
+        // Setting fixed width for checkmark because systems may have a differing width for a
+        // space and the Unicode checkmark.
+        checkmark_ = new QLabel(" ", this);
+        checkmark_->setFixedWidth(fm.width(CHECKMARK));
```
mujx: The font used here to calculate the width doesn't not have the same size as the one be used on the label. I would suggest to create a temporary font with the correct font size and use that to derive the width.

christarazi: I'm assuming you mean something like this?

```cpp
QFont checkmarkFont;
checkmarkFont.setPixelSize(conf::timeline::font::timestamp);

checkmark_->setFont(checkmarkFont);
checkmark_->setFixedWidth(QFontMetrics{checkmarkFont}.width(CHECKMARK));
...
```

mujx: Yes, but I'm not sure if the stylesheet with the font size is also needed.

christarazi: I tried it without a stylesheet and it seems to make no difference.

mujx:
Thanks again!

PR #231, opened 2018-02-07
Updated Fedora installation instructions

xvitaly:
nheko is now available from Fedora's main repositories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,6 @@ pacaur -S nheko-git
 
 #### Fedora
 ```bash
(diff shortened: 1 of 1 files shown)


mujx:
I see that nheko is available only for the latest fedora. Maybe we can specify that these instructions would work for older versions?

xvitaly:
> I see that nheko is available only for the latest fedora

It available for all supported Fedora releases: F26, F27, F28.

mujx:
Thanks!

PR #232, opened 2018-02-11
RoomList: do not blindly dereference RoomState

kaniini:
This can happen when a new room is being created from a different client and thus is created
with no known state to the local client.


mujx:
Thanks for the fix!

Could you adjust this [line](https://travis-ci.org/mujx/nheko/jobs/340002585#L1774) to fix the build?

kaniini:
done

Issue #233, opened 2018-02-15
Retry indefinetly if /sync timeouts


mujx:
Currently nheko will stop trying after 7 failed attempts and will go back to the login screen.

PR #234, opened 2018-02-17
Retry initial sync forever

Cadair:
This is an attempt at fixing #233 

I haven't been able to test more than two initial syncs locally, I hope that someone with a larger account can give it a go!

### DIFF
--- a/include/ChatPage.h
+++ b/include/ChatPage.h
@@ -193,10 +193,6 @@ private:
 
         // LMDB wrapper.
         QSharedPointer<Cache> cache_;
-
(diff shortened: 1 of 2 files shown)


mujx:
Thanks a lot for the PR!

We would like to retry indefinitely as long as the result from calling /sync is 504 (Gateway Timeout). If any other error occurs we should stop syncing and fallback to the login screen.

You could modify [this signal ](https://github.com/mujx/nheko/blob/master/src/MatrixClient.cc#L388) to return the status code and based on that you decide whether or not to continue.

PR #279, opened 2018-03-19
remove unneeded time separators from the timeline when deleting a redacted message

### LINKED ISSUE #276: The date separator is not removed together with the redacted event
If the redacted event is the only event after a date separator, the separator should also be removed.

Hawkheart:
should fix #276 :)


mujx:
Thanks a lot for the fix!

PR #294, opened 2018-04-08
Implement pressing tab to navigate auto completion

### LINKED ISSUE #287: Use tab to navigate in the auto-completion pop-up
Currently the only way to select a username is by clicking.

christarazi:
Fixes #287 

Also fixes a small bug in 7874942.

### DIFF
--- a/src/SuggestionsPopup.cpp
+++ b/src/SuggestionsPopup.cpp
@@ -107,7 +107,7 @@ SuggestionsPopup::addUsers(const QVector<SearchResult> &users)
 }
 
 void
-SuggestionsPopup::cycle()
+SuggestionsPopup::cycleThroughSuggestions()
 {
         auto item = layout_->itemAt(tab_clicks_);
         if (!item) {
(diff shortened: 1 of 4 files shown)


src/SuggestionsPopup.cpp
```diff
@@ -103,3 +105,41 @@ SuggestionsPopup::addUsers(const QVector<SearchResult> &users)
 
         resize(geometry().width(), 40 * users.size());
 }
+
+void
+SuggestionsPopup::cycleThroughSuggestions()
+{
+        auto item = layout_->itemAt(tab_clicks_);
+        if (!item) {
```
mujx: You could merge those `if (!item)` statements into one by checking if the `tab_clicks_` is greater than the number of items in the layout. Or maybe by resetting the variable if it goes beyond that number at the end of the function.

src/SuggestionsPopup.cpp
```diff
@@ -101,5 +103,41 @@ SuggestionsPopup::addUsers(const QVector<SearchResult> &users)
...
 
+        tab_clicks_ = 0; // Reset to start from the beginning of pop-up window on next invocation.
+
         resize(geometry().width(), 40 * users.size());
 }
+
+void
+SuggestionsPopup::cycleThroughSuggestions()
+{
+        tab_clicks_ %= layout_->count(); // Stay within the number of items in layout.
```
mujx: Nice oneliner!

Issue #292, opened 2018-04-04
Add replies support


xvitaly:
Please add replies support: RMB on message -> Reply.

mujx:
An implementation of this should:
 - Include the selected text or the whole message in the text input.
 - Use `formatted_body` to send the reply.

PR #309, opened 2018-04-25
changed "Redact message" to "Delete message";

F1ash:
added "Edit message" for text messages;

### DIFF
--- a/include/timeline/TimelineItem.h
+++ b/include/timeline/TimelineItem.h
@@ -159,7 +159,7 @@ private:
         QMenu *contextMenu_;
         QAction *showReadReceipts_;
         QAction *markAsRead_;
-        QAction *redactMsg_;
+        QAction *deleteMsg_;
         QAction *editMsg_;
 
         QHBoxLayout *topLayout_     = nullptr;
--- a/src/timeline/TimelineItem.cc
+++ b/src/timeline/TimelineItem.cc
@@ -52,17 +52,17 @@ TimelineItem::init()
         contextMenu_      = new QMenu(this);
         showReadReceipts_ = new QAction("Read receipts", this);
         markAsRead_       = new QAction("Mark as read", this);
-        redactMsg_        = new QAction("Delete message", this);
+        deleteMsg_        = new QAction("Delete message", this);
         contextMenu_->addAction(showReadReceipts_);
         contextMenu_->addAction(markAsRead_);
-        contextMenu_->addAction(redactMsg_);
+        contextMenu_->addAction(deleteMsg_);
 
         connect(showReadReceipts_, &QAction::triggered, this, [this]() {
                 if (!event_id_.isEmpty())
                         ChatPage::instance()->showReadReceipts(event_id_);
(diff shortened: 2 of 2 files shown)


vurpo:
I don't think this should be merged as it is. MatrixClient is meant to represent the Matrix client API, but the API has no method for editing a message. It is misleading to make a method redact a message, but call it "edit".

F1ash:
> I don't think this should be merged as it is. MatrixClient is meant to represent the Matrix client API, but the API has no method for editing a message. It is misleading to make a method redact a message, but call it "edit".
yes, it is not edit a message in usual sense... Can you advise how to name this method?

vurpo:
Because Matrix has planned to add some way to update existing events (to make e.g. reactions, edits possible), I think this feature shouldn't be added until there is some spec way to do it for real. Otherwise you end up with different Matrix clients doing it in incompatible ways.

F1ash:
Some offtop... 
Can i use for implement Quote message the format of "content" which used in [Riot](https://riot.im/):

image

vurpo:
Nheko doesn't have any support for formatted messages yet, so I'm not sure about that. Also, they are currently developing a new machine-readable format for quoting/replying to messages (which is in riot develop IIRC), so it would be better to implement support for that instead, as soon as the format is finished.

F1ash:
ok...

mujx:
@F1ash You can go ahead and implement quotes & formatted text in general. The `formatted_body` property on the event should be parsed first tho. 

The event parsing is happening in the matrix-structs library.

mujx:
Check #292

F1ash:
Is dirty probe to implement Quote:
`
...
        QStringList msg_parts_;
        QString msg_body, quoted;

        switch (ty) {
        case mtx::events::MessageType::Text:
                if ( msg.startsWith(" > ") ) {
                    msg_body = msg;
                    msg_body.remove(" > ");
                    msg_parts_ = msg_body.split("\n\n", QString::KeepEmptyParts);
                    quoted = QString("<blockquote>\n<p>%1</p>\n</blockquote>\n<p>%2</p>\n")
                            .arg(msg_parts_.first()).arg(msg_parts_.last());
                    msg_body = QString(" > %1\n\n%2")
                            .arg(msg_parts_.first()).arg(msg_parts_.last());
                    body = {{"msgtype", "m.text"}, {"body", msg_body},
                            {"formatted_body", quoted}, {"format", "org.matrix.custom.html"}};
                } else {
                    body = {{"msgtype", "m.text"}, {"body", msg}};
                };
                break;
        case mtx::events::MessageType::Emote:
...
`

works good for Nheko and Riot... But i not sure  that is correctly for new (developing) features...

mujx:
@F1ash Are you currently working on replies?

F1ash:
> Are you currently working on replies?
No? at last time i have not a time for this. I have some questions about implementation Reply widget (and layout), and recursive included quotes\replies. But here that is not easy. Have a develop channel\conference for it?

PR #318, opened 2018-05-05
Initial Russian translate

bergentroll:
I prepaired Russian ts file, but I have no idea how I can test it. I just make update-translations and make release on enveronment with $LANG=ru_RU.utf8, but the nheko is still in English.
`¯\_(ツ)_/¯`


mujx:
Thanks a lot for the translation! 

That was pretty much the way to add a new ts file but for some reason the translations don't work in general (they used to). So we have to figure out why they are not applied.

mujx:
I'm putting this on hold until the issue with the translations is resolved.

mujx:
Translations are working again. Just run `make update-translations` again to fill some missing strings.

bergentroll:
@mujx, done!
There is no every strings works for some reason, such as the line 102 in nheko_ru.ts. But most of are.

mujx:
Thanks! 

We can revisit the problematic strings. I didn't have any problem with the other languages so far.

PR #319, opened 2018-05-05
Start in tray

bergentroll:
I implemented the "Start in tray" option. Also I see more simply solution if you do not want one more switch — just start in tray while "Tray" option is enabled.

And I want you to be noticed: if I start the nheko in tray or just minimze it after start and then click on settings button or some room, I get segfault. This trouble is not with only my branch, the master is in trouble too.


PureTryOut:
> Also I see more simply solution if you do not want one more switch — just start in tray while "Tray" option is enabled.

Please don't do that. I like having the application in tray when I minimize the application, but I don't want it to go to tray on start.

PR #322, opened 2018-05-10
Implement notifications for linux

twahm:


vurpo:
This pull request adds a libnotify dependency, which in turn adds a glib dependency. On my own `notifications` branch I have implemented Linux/BSD desktop notifications without adding external dependencies, by sending and receiving directly through D-Bus (the same as libnotify does, but without a dependency).

simoniz0r:
I don't know of many people that don't have libnotify installed, honestly.  Why reinvent the wheel?

vurpo:
Mostly because the people in #nheko:matrix.org preferred not to depend on glib, and using the methods through D-Bus is as easy as using libnotify anyway. (And because I've already implemented it...)

mujx:
@twahm Sorry for not communicating this earlier. I wasn't aware that you were working on the linux notifications which is already implemented by @vurpo [here](https://github.com/vurpo/nheko/tree/notifications) but it doesn't have a PR yet. 

The problem with using libnotify, which we considered after moving away from snorenotify, is that for most linux distros it requires gtk as a dependency which doesn't make much for sense for a Qt app (having both gtk and Qt as requirements).

twahm:
Then, how could I implement this with snr, first what would be the best way to add it to cmake try this:

find_package(LibsnoreQt5 0.7.0 REQUIRED)
find_package(LibsnoreSettingsQt5 0.7.0 OPTIONAL)
target_link_libraries(nheko Snore::Libsnore Snore::LibsnoreSettings)

but cmake did not find the modules.

mujx:
The linux notifications are already implemented with QDbus in the branch I mentioned above by vurpo . We won't use snorenotify because it's not maintained (e.g doesn't compile on mac) and is quite bloated for our use case.

I'm afraid I'll have to close this one.

PR #324, opened 2018-05-11
Add support for read room access

twahm:
In Room settings you can see who can join to the room, whether it's just the inviteds, anyone or anyone including guests.

I would have liked to implement the possibility of changing it, but I am not familiar with the source code and I did not find a similar function with which I could be oriented.

### DIFF
--- a/include/Cache.h
+++ b/include/Cache.h
@@ -23,6 +23,8 @@
 #include <json.hpp>
 #include <lmdb++.h>
 #include <mtx/responses.hpp>
+#include <mtx/events/join_rules.hpp>
+using mtx::events::state::JoinRule;
 
 struct RoomMember
 {
@@ -74,6 +76,9 @@ struct RoomInfo
         bool is_invite = false;
         //! Total number of members in the room.
         int16_t member_count = 0;
+        //! Who can access to the room.
+        JoinRule join_rule = JoinRule::Public;
+        bool guest_access = false;
 };
 
 inline void
@@ -164,6 +169,9 @@ public:
 
         //! Calculate & return the name of the room.
         QString getRoomName(lmdb::txn &txn, lmdb::dbi &statesdb, lmdb::dbi &membersdb);
+        //! Get room join rules
+        JoinRule getRoomJoinRules(lmdb::txn &txn, lmdb::dbi &statesdb);
+        bool getRoomGuestAccess(lmdb::txn &txn, lmdb::dbi &statesdb);
         //! Retrieve the topic of the room if any.
         QString getRoomTopic(lmdb::txn &txn, lmdb::dbi &statesdb);
         //! Retrieve the room avatar's url if any.
--- a/include/Cache.h
+++ b/include/Cache.h
@@ -22,8 +22,8 @@
 #include <QImage>
 #include <json.hpp>
 #include <lmdb++.h>
-#include <mtx/responses.hpp>
 #include <mtx/events/join_rules.hpp>
+#include <mtx/responses.hpp>
 using mtx::events::state::JoinRule;
 
 struct RoomMember
@@ -78,7 +78,7 @@ struct RoomInfo
         int16_t member_count = 0;
         //! Who can access to the room.
         JoinRule join_rule = JoinRule::Public;
-        bool guest_access = false;
+        bool guest_access  = false;
 };
 
 inline void
--- a/include/Cache.h
+++ b/include/Cache.h
@@ -84,10 +84,12 @@ struct RoomInfo
 inline void
 to_json(json &j, const RoomInfo &info)
 {
-        j["name"]       = info.name;
-        j["topic"]      = info.topic;
-        j["avatar_url"] = info.avatar_url;
-        j["is_invite"]  = info.is_invite;
+        j["name"]         = info.name;
+        j["topic"]        = info.topic;
+        j["avatar_url"]   = info.avatar_url;
+        j["is_invite"]    = info.is_invite;
+        j["join_rule"]    = info.join_rule;
+        j["guest_access"] = info.guest_access;
 
         if (info.member_count != 0)
                 j["member_count"] = info.member_count;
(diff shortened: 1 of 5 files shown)


src/Cache.cc
```diff
@@ -805,6 +814,53 @@ Cache::getRoomName(lmdb::txn &txn, lmdb::dbi &statesdb, lmdb::dbi &membersdb)
...
+        using namespace mtx::events::state;
+
+        lmdb::val event;
+        bool res = lmdb::dbi_get(
+          txn, statesdb, lmdb::val(to_string(mtx::events::EventType::RoomGuestAccess)), event);
+
+        if (res) {
+                try {
+                        StateEvent<GuestAccess> msg = json::parse(std::string(event.data(), event.size()));
+                        if (msg.content.guest_access == AccessState::CanJoin)
```
mujx: You can just write is as `return msg.content.guest_access == AccessState::CanJoin`.

src/Cache.cc
```diff
@@ -805,6 +814,53 @@ Cache::getRoomName(lmdb::txn &txn, lmdb::dbi &statesdb, lmdb::dbi &membersdb)
...
+                                return true;
+                        else
+                                return false;
+                }
+                catch (const json::exception &e) {
+                        qWarning() << QString::fromStdString(e.what());
```
mujx: You don't need the `QString::fromStdString` method here.

src/Cache.cc
```diff
@@ -435,6 +435,8 @@ Cache::saveState(const mtx::responses::Sync &res)
                 updatedInfo.avatar_url =
                   getRoomAvatarUrl(txn, statesdb, membersdb, QString::fromStdString(room.first))
                     .toStdString();
+                updatedInfo.join_rule = getRoomJoinRules(txn, statesdb);
```
mujx: This is already saved in the states db. You can remove those lines.

src/Cache.cc
```diff
@@ -467,6 +469,8 @@ Cache::saveInvites(lmdb::txn &txn, const std::map<std::string, mtx::responses::I
                 updatedInfo.avatar_url =
                   getInviteRoomAvatarUrl(txn, statesdb, membersdb).toStdString();
                 updatedInfo.is_invite = true;
+                updatedInfo.join_rule = getRoomJoinRules(txn, statesdb);
```
mujx: Same here.

include/Cache.h
```diff
@@ -164,6 +169,9 @@ class Cache : public QObject
 
         //! Calculate & return the name of the room.
         QString getRoomName(lmdb::txn &txn, lmdb::dbi &statesdb, lmdb::dbi &membersdb);
+        //! Get room join rules
+        JoinRule getRoomJoinRules(lmdb::txn &txn, lmdb::dbi &statesdb);
```
mujx: nitpick: `getRoomJoinRule`

src/dialogs/RoomSettings.cpp
```diff
@@ -59,13 +59,59 @@ RoomSettings::RoomSettings(const QString &room_id, QWidget *parent)
...
         layout->addLayout(btnLayout);
 
         connect(cancelBtn_, &FlatButton::clicked, this, &RoomSettings::closing);
-        connect(saveBtn_, &FlatButton::clicked, this, [this]() { emit closing(); });
+        connect(saveBtn_, &FlatButton::clicked, this, &RoomSettings::save_and_close);
+}
+
+void
+RoomSettings::save_and_close() {
+        // TODO: Save access changes to the room
```
mujx: To save the changes you'll have to send a state event using the http API. 

https://matrix.org/speculator/spec/HEAD/client_server/unstable.html#get-matrix-client-r0-rooms-roomid-state-eventtype-statekey

src/Cache.cc
```diff
@@ -805,6 +814,53 @@ Cache::getRoomName(lmdb::txn &txn, lmdb::dbi &statesdb, lmdb::dbi &membersdb)
...
+        bool res = lmdb::dbi_get(
+          txn, statesdb, lmdb::val(to_string(mtx::events::EventType::RoomJoinRules)), event);
+
+        if (res) {
+                try {
+                        StateEvent<JoinRules> msg = json::parse(std::string(event.data(), event.size()));
+                        return msg.content.join_rule;
+                }
+                catch (const json::exception &e) {
+                        qWarning() << QString::fromStdString(e.what());
```
mujx: Same here for `QString::fromStdString`

src/Cache.cc
```diff
@@ -604,6 +611,8 @@ Cache::getRoomInfo(const std::vector<std::string> &rooms)
                                         RoomInfo tmp =
                                           json::parse(std::string(data.data(), data.size()));
                                         tmp.member_count = getInviteMembersDb(txn, room).size(txn);
+                                        tmp.join_rule = getRoomJoinRules(txn, statesdb);
```
mujx: Invited rooms are not viewable in nheko so we don't need those.

src/Cache.cc
```diff
@@ -584,12 +588,15 @@ Cache::getRoomInfo(const std::vector<std::string> &rooms)
...
         for (const auto &room : rooms) {
                 lmdb::val data;
+                auto statesdb  = getStatesDb(txn, room);
 
                 // Check if the room is joined.
                 if (lmdb::dbi_get(txn, roomsDb_, lmdb::val(room), data)) {
                         try {
                                 RoomInfo tmp = json::parse(std::string(data.data(), data.size()));
                                 tmp.member_count = getMembersDb(txn, room).size(txn);
+                                tmp.join_rule = getRoomJoinRules(txn, statesdb);
```
mujx: Could you copy those lines in `Cache::singleRoomInfo` also, and make the `RoomSettings` page call that method instead of `Cache::getRoomInfo`.

src/dialogs/RoomSettings.cpp
```diff
@@ -59,13 +58,59 @@ RoomSettings::RoomSettings(const QString &room_id, QWidget *parent)
         notifOptionLayout_->addWidget(notifLabel);
         notifOptionLayout_->addWidget(notifCombo, 0, Qt::AlignBottom | Qt::AlignRight);
 
+        auto accessOptionLayout = new QHBoxLayout();
+        accessOptionLayout->setMargin(5);
+        auto accessLabel = new QLabel(tr("Room access"), this);
+        accessCombo = new QComboBox(this);
+        accessCombo->addItem(tr("Anyone and guests"));
+        accessCombo->addItem(tr("Anyone"));
+        accessCombo->addItem(tr("Inviteds"));
```
mujx: `Invited users`

src/dialogs/RoomSettings.cpp
```diff
@@ -59,13 +58,59 @@ RoomSettings::RoomSettings(const QString &room_id, QWidget *parent)
...
+        accessCombo->setDisabled(true);
+        accessLabel->setStyleSheet("font-size: 15px;");
+
+        if(info_.join_rule == JoinRule::Public)
```
mujx: Could you add some braces around the outer `if` statement.

PR #333, opened 2018-06-10
slightly improved ru+de translations

se17p:
Please review.
Not sure about what ` type="unfinished"` means and if it should be removed though.


mujx:
The `type="unfinished` it's mostly used by the Qt LInguist tool. You can remove those if there is a translation.

Unfortunately I don't know Russian or German so somebody else will review.

rokups:
Russian looks good. I would expect "пишет" instead of "печатает" though, but im no native speaker so i may be wrong.

se17p:
removed type=unfinished for translated items

Issue #338, opened 2018-06-18
build fails with bundled boost when system has older version from distro repository


4nd3r:
this is actually two separate issues, which are somewhat related, hence one issue for both.

first: newsest available boost in Debian Sid is 1.63 (https://packages.debian.org/source/sid/boost1.63).

second: when using `-DUSE_BUNDLED_BOOST=ON`, build fails with MatrixClient:

```
$ cmake -Hdeps -B.deps -DUSE_BUNDLED_BOOST=ON -DUSE_BUNDLED_SPDLOG=ON
--8<--
$ cmake --build .deps
[ 20%] Built target Boost
[ 22%] Performing update step for 'Olm'
[ 25%] Performing configure step for 'Olm'
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ander/Source/nheko/.deps/build/olm
[ 27%] Performing build step for 'Olm'
[100%] Built target olm
[ 30%] Performing install step for 'Olm'
[100%] Built target olm
Install the project...
-- Install configuration: "Release"
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/lib/libolm.a
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/include/olm/olm.h
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/include/olm/outbound_group_session.h
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/include/olm/inbound_group_session.h
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/lib/cmake/Olm/OlmTargets.cmake
-- Installing: /home/ander/Source/nheko/.deps/usr/lib/cmake/Olm/OlmTargets-release.cmake
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/lib/cmake/Olm/OlmConfig.cmake
-- Up-to-date: /home/ander/Source/nheko/.deps/usr/lib/cmake/Olm/OlmConfigVersion.cmake
[ 32%] Completed 'Olm'
[ 40%] Built target Olm
[ 42%] Performing update step for 'SpdLog'
[ 45%] Performing configure step for 'SpdLog'
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ander/Source/nheko/.deps/build/spdlog
[ 47%] Performing build step for 'SpdLog'
[ 50%] Performing install step for 'SpdLog'
Install the project...
-- Install configuration: ""
[...]

PR #350, opened 2018-06-22
Build mtxclient with Bundled Boost

camconn:
This commit allows systems with outdated versions of Boost to build `mtxclient` with the bundled version of Boost when `USE_BUNDLED_BOOST` is specified. Without this, the build for `mtxclient` will fail.

This PR resolves mujx/nheko#338


deps/cmake/MatrixClient.cmake
```diff
@@ -8,6 +9,13 @@ if(APPLE)
     set(PLATFORM_FLAGS "-DOPENSSL_ROOT_DIR=/usr/local/opt/openssl")
 endif()
 
+# Force to build with the bundled version of Boost, if requested. This is
+# necessary because if an outdated Boost is installed, then CMake will grab
+# that instead of the bundled version of Boost, like we wanted.
+if(USE_BUNDLED_BOOST)
```
mujx: You could remove this `if` statement because this file will only be included if that flag is `true` anyway.

camconn:
Thanks for the feedback. I removed the `if` statement.

Issue #348, opened 2018-06-21
E2EE: Image from clipboard data / upload is sent unencrypted


lbeltrame:
### System:

- Nheko version: 0.4.3-74396e9
- Installation method: Source
- Operating System: openSUSE Tumbleweed
- Qt version: 5.11
- C++ compiler: GCC 8.1
- Desktop Environment: Plasma 5.13.80

### Actual behavior

Pasting an image from the clipboard will cause the attached image to be sent unencrypted.  Uploading the image via the attachment icon does the same.
Checking the raw source of the image shows indeed that it was sent without being encrypted.

### Expected behavior

Image is sent encrypted.

### Steps to reproduce

1. Access an E2EE-enabled room
2. Paste / attach an image so that it is shown inline
3. Check from Riot the state of the message

albjeremias:
user content is tottaly broken on riot... 

checkout this issue: https://github.com/vector-im/riot-web/issues/6173#issuecomment-366040915 

here some comments on the code:
https://github.com/matrix-org/matrix-react-sdk/blob/master/src/components/views/messages/MFileBody.js#L66-L116

mujx:
Encrypted attachments are not implemented yet.

lbeltrame:
Can the attachment be disabled (only if it doesn't require workarounds) in E2EE rooms? Right now there's no way to tell from nheko that they're sent unencrypted.

mujx:
It would be better IMHO to wait for proper support than trying to add now an option to disable them which most likely be removed in the future.

lbeltrame:
Sounds reasonable to me. FTR, I'm perfectly fine with this feature not being present yet, but there should be at least a warning somewhere (and IMO the README.md is the best place for it) that not all content is encrypted (or conversely, indicating what is supported, like "attachments not encrypted yet").

If you think it makes sense, I can make a PR against the README.md.

PR #352, opened 2018-06-22
State that encrypted attachments aren't supported

lbeltrame:
As nheko does not support yet sending encrypted attachments (see issue #348) explicitly indicate so in the README.

(NOTE: this was made with GH's interface.... if it looks messy, I'll just do it "the usual way")


mujx:
Looks good. It just needs to be cleaned up a bit to not have such a large diff.

lbeltrame:
I blame the GH "editor". I'll properly fork and do a "real" diff today. Sorry for the inconvenience.

lbeltrame:
@mujx Ok, now it should be done without all those spurious changes.

PR #360, opened 2018-06-28
D-Bus desktop notifications

vurpo:
Adds desktop notifications on systems which have a notification daemon listening on D-Bus at `org.freedesktop.Notifications` according to the Freedesktop notifications spec (this means most Linux and BSD desktop OSes). Notifications can be clicked to open the corresponding room.


vurpo:
I broke a lot of things in my local repo, I'll be back in a moment

PR #361, opened 2018-06-28
Working D-Bus desktop notifications

vurpo:
I'm back! This PR adds desktop notifications on systems which have a notification daemon listening on D-Bus at org.freedesktop.Notifications according to the Freedesktop notifications spec (this means most Linux and BSD desktop OSes). Notifications can be clicked to open the corresponding room.


mujx:
The errors are related to the slots not being defined for windows & macos. Plus the missing curly braces from the constructor in macos.

mujx:
Thanks a lot!!

PR #366, opened 2018-07-09
Fixed warnings introduced by -Wshadow

Tomatower:
Commit 172d111d1cdb8414da9a3a81b2626db2b14c9e5e introduced loads of `-Wshadow -Werror` - errors, this PR should fix them.

### DIFF
--- a/src/timeline/TimelineView.cc
+++ b/src/timeline/TimelineView.cc
@@ -293,8 +293,8 @@ TimelineView::parseEncryptedEvent(const mtx::events::EncryptedEvent<mtx::events:
                         // TODO: request megolm session_id & session_key from the sender.
                         return {dummy, false};
                 }
-        } catch (const lmdb::error &e) {
-                nhlog::db()->critical("failed to check megolm session's existence: {}", e.what());
+        } catch (const lmdb::error &err) {
+                nhlog::db()->critical("failed to check megolm session's existence: {}", err.what());
                 dummy.content.body = "-- Decryption Error (failed to communicate with DB) --";
                 return {dummy, false};
         }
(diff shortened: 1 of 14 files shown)


mujx:
Thanks for the patch!

From a closer look most of the errors are false positives (most of them are regarding lambdas). What compiler are you using? I would suggest removing the flag for that specific compiler.

Clang doesn't have that problem.

Tomatower:
My compiler is `gcc (Gentoo 7.3.0-r3 p1.4) 7.3.0`

Yap - all fixed ones where false positives. But if you are building with -Werror, then every single warning should be fixed.

mujx:
Given that this is a bug in GCC it would better to toggle the `-Wshadow` flag off for that specific compiler.

Tomatower:
Yes and no - in a lot of places (like in constructors) it is intended behaviour to shade over variable names, but regarding lambdas read : https://stackoverflow.com/a/42089831/1080604 - clang might soon follow.

mujx:
All of the lambdas in the diff are false positives tho, because the "shadowed" parameter isn't captured.

src/timeline/TimelineView.cc
```diff
@@ -304,22 +304,22 @@ TimelineView::parseEncryptedEvent(const mtx::events::EncryptedEvent<mtx::events:
...
                                       index.room_id,
                                       index.session_id,
                                       index.sender_key,
-                                      e.what());
+                                      err.what());
                 dummy.content.body =
                   "-- Decryption Error (failed to retrieve megolm keys from db) --";
                 return {dummy, false};
-        } catch (const mtx::crypto::olm_exception &e) {
+        } catch (const mtx::crypto::olm_exception &err) {
```
mujx: Those are the only valid changes I see.

mujx:
Also related bug report https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79328. The third example in the attachment (https://gcc.gnu.org/bugzilla/attachment.cgi?id=40647) is our case.

PR #368, opened 2018-07-10
limit -Wshadow to non-gcc

Tomatower:
Due to Bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79328, limit -Wshadow to non-gcc compilers (also see #366 )


mujx:
Thanks!

PR #370, opened 2018-07-12
Added FindOlm.cmake script from mtxclient

xvitaly:
Added FindOlm.cmake script from mtxclient to find generic libolm packaged without cmake support.


mujx:
Thanks! I have forgotten about it.

PR #395, opened 2018-07-31
Add install instructions for macOS

scarface-one:
macports does either source based or binary installs. Its currently working 10.12 and above, but fails to build on older versions.


mujx:
Thanks a lot for uploading it to macports!

Could you adjust the diff so it only shows what changed? I suppose this was done by the Github UI.

mujx:
Yes please, that would be best. It's possible to use the same PR though.

scarface-one:
@mujx the diff is fixed now

Issue #80, opened 2017-09-30
Implement grouping functionality for rooms


hurzelmann:
As a user of nheko I would like to group chats into several sections, like "Private", "Business", "Project A", "Project B", etc.

vberger:
NB: This could/should be done using matrix tags, (cc https://github.com/matrix-org/matrix-doc/issues/931 which unfortunately has not received much attention) in order to all categories to be available cross-clients.

PureTryOut:
I'm assuming this is different from the upcoming group support in Riot and Synapse? If so, the current title is confusing. Maybe "Implement tagging functionality for rooms" or the likes would be better?

hurzelmann:
renamed

rodneyrod:
It looks like Quaternion has already pulled in this feature, is there any possibility of using their code as a basis for this?

PR #401, opened 2018-08-09
Display tags as sorting items in the community panel

elinorbgr:
This PR adds the basics or room tagging functionality, for #80.

Currently, it only adds a `tags` field to `RoomInfo` containing the list of associated tags, discarding any metadata they convey. I would like some feedback about whether this is the correct route for implementing this state tracking before inserting more elaborate logic in there.


mujx:
:+1: 

This seems like the correct path, saving the necessary info in cache between sync requests.

Valodim:
Providing motivation: "Pinning" of rooms the way Riot does with Favorites (which uses room tags) is the one thing I'm missing from nheko to use it as a daily driver these days. Very excited for this feature :)

vberger:
I tried rebasing this before continuing to advance, and the compiler now complains with:

```
.../nheko/src/Cache.cpp: In member function ‘void Cache::saveState(const mtx::responses::Sync&)’:
.../nheko/src/Cache.cpp:872:29: error: ‘mpark’ has not been declared
                         if (mpark::holds_alternative<
                             ^~~~~
```

Has something been changed recently regarding the way dependencies are handled... ?

mujx:
Yes the `mpark` library has been removed, the API is slightly different.

vberger:
Okay, here is basic tag support. A generic "tag" icon is used for tags in the community tab.

I think that, for a minimally ergonomic UI, this needs to be added:

- [ ] display tag name as a tooltip on pointer overlay
- [ ] use custom icons for `m.favourite` and `m.lowpriority`
- [ ] ensure special ordering for tags and communities
  - `world` remains first
  - `m.favourite` is second
  - communities and tags are here, in an order to be decided
  - `m.lowpriority` is last

vberger:
I have setup basic icons for favourite and lowpriority rooms. They are probably not the best possible, but improving them is just a matter of changing the PNG files, and I guess this wouldn't block this PR?

Regarding the tooltip, I've gone for the convention:

- "All rooms" / "Favourite rooms" / "Low priority rooms" for world/m.favourite/m.lowpriority
- "<tag name> (tag)" for tags
- "<community name> (community)" for communities

Does that sound good?

Regarding the last point, I don't know how it is possible to manually order the elements in the vertical bar. Any hints somebody?

mujx:
I've just tested the favourite & low priority tags and it works great! Not sure how we can test custom tags. Is there any other way than manually making the API calls?

The icons are not a blocker as they could be easily changed if something better comes up.

Sorting will have to be done manually (removing & re-positioning the widgets in the layout) as far as I know. There is something similar [here](https://github.com/mujx/nheko/blob/master/src/RoomList.cpp#L294).

anoadragon453:
Allowing for creation of room tags doesn't seem that tricky, and would be a major win over Riot :)

vberger:
> Is there any other way than manually making the API calls?

Unfortunately, not afaik. I'm also in favour of adding such an interface to nheko, but I'll reserve that for a future PR of that's okay.

For now, it's possible to add a tag to a room with the following curl command:

```
curl -X PUT -d '{"order": "1"}' "https://<homeserver>/_matrix/client/r0/user/<your-username>/rooms/<the-room-id>/tags/<chosen-tag>?access_token=<access_token>"
```

You need to fill in:

- `<homeserver>` : your homeserver domain
- `<your-username>` : your full username (`@login:homeserver.tld`)
- `<the-room-id>` : the id (starting with `!`) of the room to tag, aliases do not work
- `<chosen-tag>` : the tag to add to the room. Spec conventions user tags to be any utf8 string prefixed by `u.`.
- `<access_token>` : an access token

vberger:
Woops, looks like [`QSharedPointer::get`](http://doc.qt.io/qt-5/qsharedpointer.html#get) is too new for the Qt version of the CI, I'll replace that with [`QSharedPointer::data`](http://doc.qt.io/qt-5/qsharedpointer.html#data).

src/CommunitiesList.cpp
```diff
@@ -56,6 +64,59 @@ CommunitiesList::setCommunities(const mtx::responses::JoinedGroups &response)
...
+                    tag.compare(0, 2, "m."))
+                        continue;
+                QString name = QString("tag:") + QString::fromStdString(tag);
+                if (!communityExists(name)) {
+                        addCommunity(std::string("tag:") + tag);
+                }
+        }
+        // update membership of the room for all tags
+        auto it = communities_.begin();
+        while (it != communities_.end()) {
```
mujx: Maybe I'm missing something but can't this be translated into a standard for-range loop?

src/CommunitiesListItem.cpp
```diff
@@ -98,11 +113,33 @@ CommunitiesListItem::resolveName() const
...
         return firstPart.right(firstPart.size() - 1);
 }
+
+void
+CommunitiesListItem::updateTooltip()
+{
+        if (groupId_ == "world")
+                setToolTip(tr("All rooms"));
+        else if (is_tag()) {
+                QString tag = groupId_.right(groupId_.size() - 4);
```
mujx: It would be nice to have this `4` number derived from the `tag:` string somehow. Currently it seems like a magic number if you only read a part of the code.

mujx (changes requested):
Thanks again for your work.

I've highlighted some small refactorings

src/CommunitiesList.cpp
vberger: I don't know how to do the `it = communities_.erase(it);` (at like 115) with a standard for-range loop, is that possible?

src/CommunitiesListItem.cpp
vberger: Is adding a comment explaining it enough, or should I replace it by `strlen("tag:")`?

src/CommunitiesList.cpp
```diff
@@ -56,6 +64,59 @@ CommunitiesList::setCommunities(const mtx::responses::JoinedGroups &response)
...
+{
+        // create missing tag if any
+        for (const auto &tag : tags) {
+                // filter out tags we should ignore according to the spec
+                // https://matrix.org/docs/spec/client_server/r0.4.0.html#id154
+                // nheko currently does not make use of internal tags
+                // so we ignore any tag containig a `.` (which would indicate a tag
+                // in the form `tld.domain.*`) except for `m.*` and `u.*`.
+                if (tag.find(".") != ::std::string::npos && tag.compare(0, 2, "m.") &&
+                    tag.compare(0, 2, "m."))
```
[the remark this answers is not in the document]
vberger: Indeed I'll correct that.

src/CommunitiesList.cpp
```diff
@@ -193,3 +254,40 @@ CommunitiesList::roomList(const QString &id) const
...
+                        else if (entry.first == "tag:m.lowpriority")
+                                footer.push_back(item);
+                        else
+                                tags.push_back(item);
+                else
+                        communities.push_back(item);
+        }
+
+        contentsLayout_->insertWidget(contentsLayout_->count() - 1, communities_["world"].data());
+        for (auto item : header)
```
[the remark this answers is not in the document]
vberger: Are you referring to the fact that `contentsLayout_->insertWidget(contentsLayout_->count() - 1, item);` is a long call and that it implies a lot of repetition?

I realized I could simply remove the final "Stretch" item and insert it back in the layout at the end, rather than insert all widgets "one before the end". Would that be okay too?

src/ChatPage.cpp
```diff
@@ -1079,6 +1081,9 @@ ChatPage::trySync()
                           emit syncTopBar(updates);
                           emit syncRoomlist(updates);
 
+                          auto tag_updates = cache::client()->roomTagUpdates(res);
+                          emit syncTags(tag_updates);
```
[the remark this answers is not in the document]
vberger: Will do.

src/CommunitiesList.cpp
mujx: You're right, I missed that part. You would need the iterator.

src/CommunitiesListItem.cpp
mujx: I've seen it in 2-3 places so maybe it would be better to use something like `strlen`.

src/CommunitiesList.cpp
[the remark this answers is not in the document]
mujx: Yes, I was referring to that repetition. I'm fine with any solution that doesn't have a lot of duplication.

Valodim:
(with apologies for commenting from the sidelines:) If you're going to do custom room tags in the future, please write up a brief concept document and think through the user workflows first, and ideally sync up with Riot and their possible plans. It doesn't *seem* tricky because the technical bits are easy, but "going with the code flow" on these kinds of features is a slippery slope towards a fragmented ecosystem and bad user experience.

Thanks for your work either way, really looking forward to the release that includes this!

vberger:
@Valodim Regarding that, I had already pushed a few month ago for a better specification of the semantics of tags in the client-server spec: https://matrix.org/docs/spec/client_server/r0.4.0.html#id154, I believe this already sets a good basis for relative uniformity, do you think more is needed?

mujx:
@Valodim Not sure what you mean by code flow?

Custom tags are supported by the spec so we're just taking advantage of an otherwise under-appreciated feature.

vberger:
@mujx I believe this last commit should address your comments.

Valodim:
Oh, I didn't know there were namespaces for tags yet! Yeah that should work nicely, and great to hear there is initiative wrt spec work, too.

Relating to my previous comment, this seems like a use case that other clients may want to support. A vendor-neutral namespace with well defined semantics for the workflows you want to cover would facilitate that.

Perhaps the `u.` namespace would be sufficient for a straightforward first iteration, actually :)

PR #402, opened 2018-08-09
kill me please for this (polish translation)

m4sk1n:


mujx:
So it was Czech initially?

m4sk1n:
@mujx nope, it was in Polish, but I used less suited word in this case and because I have copied it from original, just because it's similar, I forgot to change one character, so I made spelling error

PR #403, opened 2018-08-11
Enable compilation on FreeBSD

dramborleg:
These are the (very simple) changes I made to allow me to compile and use Nheko on FreeBSD. Please let me know what you think, if I need to do additional modifications I'd be happy to give it a shot as time allows.


mujx:
Thanks a lot for making it usable in FreeBSD!

dramborleg:
haha all 2 lines of code! Thanks for building this, it's great to be able to use it.

PR #448, opened 2018-09-24
Fix up includes to nlohmann json

z3ntu:
Might require more changes for the changed include path

See also https://github.com/mujx/mtxclient/pull/29


z3ntu:
The build fails because the mtxclient PR isn't used in the build process.

deps/CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ set(JSON_HEADER_URL
 set(JSON_HEADER_HASH
     ce6b5610a051ec6795fa11c33854abebb086f0fd67c311f5921c3c07f9531b44)
 
-file(DOWNLOAD ${JSON_HEADER_URL} ${DEPS_INSTALL_DIR}/include/json.hpp
+file(DOWNLOAD ${JSON_HEADER_URL} ${DEPS_INSTALL_DIR}/include/nlohmann/json.hpp
```
mujx: It would be better to use `External_Project` for this library and install it like the other dependencies.  And then call `find_package(nlohmann_json)` which will fix the CI.

z3ntu: Probably but it's more work :wink: I'll look into it soon-ish

